PR #4, opened 2024-08-20
Initial device reporting in config daemon

almaslennikov:


pkg/host/utils.go
```diff
@@ -0,0 +1,184 @@
...
+
+	"github.com/Mellanox/nic-configuration-operator/pkg/consts"
+	dputils "github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/utils"
+	"github.com/vishvananda/netlink"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const PciDevicesPath = "/sys/bus/pci/devices"
+
+type HostUtils interface {
```
ykulazhenkov, line 19: Can you add comments to the Interface and functions?

pkg/daemon/daemon.go
```diff
@@ -1 +1,107 @@
...
+
+	"github.com/Mellanox/nic-configuration-operator/api/v1alpha1"
+	"github.com/Mellanox/nic-configuration-operator/pkg/host"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+type NicConfigurationDaemon struct {
```
ykulazhenkov, line 16: Can you add comments to the struct and functions?

ykulazhenkov (commented):
partial review

config/daemon/daemon.yaml
```diff
@@ -17,9 +17,13 @@ spec:
...
         control-plane: nic-configuration-daemon
     spec:
+      hostNetwork: true
+      hostPID: true
+      hostIPC: true
+      priorityClassName: system-node-critical
       containers:
-        - image: nic-configuration-daemon:latest
-          imagePullPolicy: Never
+        - image: harbor.mellanox.com/cloud-orchestration-dev/nic-configuration-daemon:latest
```
adrianchiris, line 25: lets keep original value here.

adrianchiris: this customizations should be made with kustomize when deploying

config/daemon/daemon.yaml
```diff
@@ -17,9 +17,13 @@ spec:
...
+          #imagePullPolicy: Never
```
adrianchiris, line 26: remove altogether if its not needed.

go.mod
```diff
@@ -3,22 +3,28 @@ module github.com/Mellanox/nic-configuration-operator
 go 1.21
 
 require (
+	github.com/jaypipes/ghw v0.12.0
+	github.com/k8snetworkplumbingwg/sriov-network-device-plugin v3.0.0+incompatible
 	github.com/onsi/ginkgo/v2 v2.14.0
 	github.com/onsi/gomega v1.30.0
+	github.com/vishvananda/netlink v1.1.0
```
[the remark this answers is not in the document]
adrianchiris: +1

go.mod
```diff
@@ -3,22 +3,28 @@ module github.com/Mellanox/nic-configuration-operator
 go 1.21
 
 require (
+	github.com/jaypipes/ghw v0.12.0
+	github.com/k8snetworkplumbingwg/sriov-network-device-plugin v3.0.0+incompatible
```
adrianchiris, line 7: should we just shamelessly copy the dependencies ? we dont claim to have stable API for pkgs in device plugin.

pkg/log/log.go
```diff
@@ -0,0 +1,54 @@
+package ncolog
```
adrianchiris, line 1: usually pkg name matches the folder.

why not call this one log ?

adrianchiris: also do you need this pkg ?

main reason to have it is to support runtime updates of log level.

that said its fine to have it for intialization of zap logger.

cmd/nic-configuration-daemon/main.go
```diff
@@ -1,5 +1,54 @@
...
+		os.Exit(1)
+	}
+
+	// Get the pod name and namespace from the environment variables
+	nodeName := os.Getenv("NODE_NAME")
+
+	hostUtils := host.NewHostUtils()
+	configDaemon := daemon.New(kubeClient, nodeName, host.NewHostManager(nodeName, hostUtils), hostUtils)
 
+	err = configDaemon.Run()
```
adrianchiris, line 50: what about signal handling ?

cmd/nic-configuration-daemon/main.go
```diff
@@ -1,5 +1,54 @@
...
+	if err != nil {
+		log.Log.Error(err, "failed to create clientset")
+		os.Exit(1)
+	}
+
+	// Get the pod name and namespace from the environment variables
+	nodeName := os.Getenv("NODE_NAME")
+
+	hostUtils := host.NewHostUtils()
+	configDaemon := daemon.New(kubeClient, nodeName, host.NewHostManager(nodeName, hostUtils), hostUtils)
```
adrianchiris, line 48: daemon is going to have several parts ("controllers")
what do you think about using controller manager for it ?

you can register reconcilers as well as add "runnable" components if needed.

pkg/daemon/daemon.go
```diff
@@ -1 +1,107 @@
...
+
+		err = d.UpdateDeviceStatus(device, deviceStatus)
+		if err != nil {
+			log.Log.Error(err, "Failed to update NicDevice status")
+			continue
+		}
+	}
+
+	time.Sleep(time.Second * 1000)
+	return nil
```
adrianchiris, line 59: so its one shot ? shouldnt it run periodically ?

adrianchiris: what if error happens on one device ? when would it be retried ?

pkg/daemon/daemon.go
```diff
@@ -1 +1,107 @@
...
+
+	hostManager := host.NewHostManager(d.nodeName, hostUtils)
+
+	devices := hostManager.DiscoverNicDevices()
+
+	for _, deviceStatus := range devices {
+		log.Log.Info("Device found", "deviceStatus", deviceStatus)
+
+		name := deviceStatus.Type + "-" + strings.ToLower(deviceStatus.SerialNumber)
+		namespace := "default"
```
adrianchiris, line 43: this should be the operator namespace no ?

pkg/daemon/daemon.go
```diff
@@ -1 +1,107 @@
...
+		}
+	}
+
+	time.Sleep(time.Second * 1000)
+	return nil
+}
+
+func (d *NicConfigurationDaemon) createDeviceCRIfNotExists(name string, namespace string) (*v1alpha1.NicDevice, error) {
+	existingDevice := &v1alpha1.NicDevice{}
+	err := d.client.Get(context.TODO(), client.ObjectKey{
```
adrianchiris, line 64: we do want some valid context here : e.g a context that is cancelled if signal recieved.

pkg/host/utils.go
```diff
@@ -0,0 +1,184 @@
+package host
```
adrianchiris, line 1: please add license headers for everything

pkg/host/utils.go
```diff
@@ -0,0 +1,184 @@
...
+		return "", "", err
+	}
+
+	// Parse the output for PN and SN
+	scanner := bufio.NewScanner(strings.NewReader(string(output)))
+	var partNumber, serialNumber string
+
+	for scanner.Scan() {
+		line := scanner.Text()
+
```
adrianchiris, line 68: do we want "ToLower" for these strings ?

pkg/host/utils.go
```diff
@@ -0,0 +1,184 @@
...
+			serialNumber = strings.TrimSpace(strings.TrimPrefix(line, "SN:"))
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		log.Log.Error(err, "Error reading mstvpd output")
+		return "", "", err
+	}
+
+	return partNumber, serialNumber, nil
```
adrianchiris, line 82: what if one found but not the other ?

pkg/host/utils.go
```diff
@@ -0,0 +1,184 @@
...
+	scanner := bufio.NewScanner(strings.NewReader(string(output)))
+	var firmwareVersion, PSID string
+
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if strings.HasPrefix(line, "FW Version:") {
+			firmwareVersion = strings.TrimSpace(strings.TrimPrefix(line, "FW Version:"))
+		}
+		if strings.HasPrefix(line, "PSID:") {
```
adrianchiris, line 104: same WRT the "ToLower" comment

pkg/host/utils.go
```diff
@@ -0,0 +1,184 @@
...
+			PSID = strings.TrimSpace(strings.TrimPrefix(line, "PSID:"))
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		log.Log.Error(err, "Error reading mstflint output")
+		return "", "", err
+	}
+
+	return firmwareVersion, PSID, nil
```
adrianchiris, line 114: nit: ensure both are non-empty

pkg/host/utils.go
```diff
@@ -0,0 +1,184 @@
...
+	totalVfFilePath := filepath.Join(PciDevicesPath, pciAddr, "physfn")
+	if _, err := os.Stat(totalVfFilePath); err != nil {
+		return false
+	}
+	return true
+}
+
+func (h *hostUtils) GetRDMADeviceName(pciAddr string) string {
+	log.Log.Info("HostUtils.GetRDMADeviceName()", "pciAddr", pciAddr)
+
```
adrianchiris, line 160: can we leverage rdmamap lib for it ?

https://github.com/Mellanox/rdmamap/blob/d8230af2dc589a7e0a421e7b9927284d838d878a/rdma_map.go#L386

pkg/host/host.go
```diff
@@ -0,0 +1,95 @@
...
+
+			devices[deviceStatus.SerialNumber] = deviceStatus
+		}
+	}
+
+	return devices
+}
+
+func (h hostManager) UpdateDeviceConditionsStatus(device *v1alpha1.NicDeviceStatus) {
+	device.Conditions = []v1alpha1.NicDeviceConditionSpec{{
```
adrianchiris, line 87: there is a lib in api machinery to deal with conditions. i strongly suggest to use it.

as an example. you must set last transition time, which it will take care for u.
moreover if there are other conditions this function will remove them and replace with this single condition.

adrianchiris (commented):
what are the plans for unit-tests ?

go.mod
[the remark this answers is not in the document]
almaslennikov: Changed it to a recently released v1.3.0

pkg/consts/consts.go
```diff
@@ -0,0 +1,13 @@
+package consts
+
+const (
+	MELLANOX_VENDOR = "15b3"
```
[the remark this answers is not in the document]
almaslennikov: Done for all consts

go.mod
almaslennikov: Copied the function locally

pkg/log/log.go
almaslennikov: Changed the folder name to match the package

cmd/nic-configuration-daemon/main.go
almaslennikov: Changed daemon to be a NicDevice reconciler, now it uses a controller manager

pkg/daemon/daemon.go
almaslennikov: Resolved by using controller manager

pkg/host/utils.go
almaslennikov: Now it returns an error

pkg/host/utils.go
almaslennikov: Done

pkg/host/host.go
almaslennikov: Done, the logic will evolve, now added only a basic condition

Dockerfile.nic-configuration-daemon
```diff
@@ -29,7 +29,6 @@ RUN --mount=type=cache,target=/go/pkg/mod/ GO_GCFLAGS=${GCFLAGS} make build-daem
 FROM quay.io/centos/centos:stream9
```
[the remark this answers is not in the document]
almaslennikov: Not really, is there a reason not to?

almaslennikov:
@adrianchiris @ykulazhenkov addressed your comments for the current review iteration.
the unit tests are to follow

Dockerfile.nic-configuration-daemon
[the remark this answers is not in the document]
ykulazhenkov: I mean currently you don't install anything to the image and I don't see why you need to switch from the distrolless. Do we need any OS packages in the container?

pkg/host/host.go
```diff
@@ -0,0 +1,126 @@
...
+type hostManager struct {
+	nodeName  string
+	hostUtils HostUtils
+}
+
+// DiscoverNicDevices uses host utils to discover Nvidia NIC devices on the host and returns back a map of serial numbers to device statuses
+func (h hostManager) DiscoverNicDevices() map[string]v1alpha1.NicDeviceStatus {
+	log.Log.Info("HostManager.DiscoverNicDevices()")
+
+	unchroot, err := h.hostUtils.Chroot("/host")
```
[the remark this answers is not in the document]
almaslennikov: We are executing a lot of system tools, where it's not always entirely clear, which system directories they try to access. If needed, we can try to make do with a handful of specific directories mounts but that's subject to problems in the future IMHO

internal/controller/nicdevice_controller.go
```diff
@@ -0,0 +1,203 @@
...
+		GenericFunc: func(e event.GenericEvent) bool {
+			obj, ok := e.Object.(*v1alpha1.NicDevice)
+			if !ok {
+				return false
+			}
+			return obj.Status.Node == r.nodeName
+		},
+	}
+
+	go func() {
```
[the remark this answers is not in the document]
almaslennikov: Changed the design to have two runnables as part of controller manager:
1. Device registry which is a simple runnable executing every 5 minutes. It discovers the devices and creates/deletes/updates the status if needed
2. NicDevice CR reconciler. It's triggered by spec updates and will apply the requested configuration and conditions

internal/controller/nicdevice_controller.go
```diff
@@ -0,0 +1,203 @@
...
+
+//+kubebuilder:rbac:groups=configuration.net.nvidia.com,resources=nicdevices,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=configuration.net.nvidia.com,resources=nicdevices/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=configuration.net.nvidia.com,resources=nicdevices/finalizers,verbs=update
+
+func (r *NicDeviceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = log.FromContext(ctx)
+	devices := r.hostManager.DiscoverNicDevices()
+
+	for _, deviceStatus := range devices {
```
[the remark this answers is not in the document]
almaslennikov: Good idea, thanks

internal/controller/nicdevice_controller.go
```diff
@@ -0,0 +1,203 @@
...
+		log.Log.Info("Device found", "deviceStatus", deviceStatus)
+
+		name := deviceStatus.Type + "-" + strings.ToLower(deviceStatus.SerialNumber)
+
+		device, err := r.createDeviceCRIfNotExists(ctx, name, r.namespace)
```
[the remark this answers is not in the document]
almaslennikov: See comment above

pkg/host/utils.go
```diff
@@ -0,0 +1,226 @@
...
+	"strings"
+	"syscall"
+
+	"github.com/Mellanox/nic-configuration-operator/pkg/consts"
+	"github.com/Mellanox/rdmamap"
+	"github.com/vishvananda/netlink"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const PciDevicesPath = "/sys/bus/pci/devices"
```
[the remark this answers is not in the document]
almaslennikov: changed to start with lower case

internal/controller/nicdevice_controller.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// UpdateDeviceStatus calculates the status conditions and updates the Status field of the CR
+func (r *NicDeviceReconciler) UpdateDeviceStatus(ctx context.Context, device *v1alpha1.NicDevice) error {
+	r.hostManager.UpdateDeviceConditionsStatus(device)
+
+	err := r.Client.Status().Update(ctx, device)
```
ykulazhenkov, line 75: Should you call the status update only in case if the status field was modified?

internal/controller/nicdevice_controller.go
```diff
@@ -0,0 +1,144 @@
...
+	err := r.Get(ctx, req.NamespacedName, device)
+	if err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
+
+	if device.Spec.Configuration == nil {
+		return reconcile.Result{}, nil
+	}
+
+	err = r.UpdateDeviceStatus(ctx, device)
```
ykulazhenkov, line 62: Currently, If the spec is empty the status will not be updated with the "spec not set" condition. Is this intentional?

pkg/host/host.go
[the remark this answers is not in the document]
ykulazhenkov: If I understand correctly, this will case that we will call `mstconf` and `mstvpd` from the host (meaning these tools should be instaleld on the host OS) . I think this is not what we want.

config/daemon/daemon.yaml
```diff
@@ -17,9 +17,12 @@ spec:
       labels:
         control-plane: nic-configuration-daemon
     spec:
+      hostNetwork: true
+      hostPID: true
+      hostIPC: true
```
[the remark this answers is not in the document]
almaslennikov: No, we don't. Removed this

internal/controller/nicdevice_controller.go
almaslennikov: Reworked the status conditions. No they are only set in the controller (apart from the initial condition)

Dockerfile.nic-configuration-daemon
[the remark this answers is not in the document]
almaslennikov: Let's use distroless for now and switch later if needed

pkg/host/host.go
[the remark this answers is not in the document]
almaslennikov: @ykulazhenkov Even if we install those tools inside the container, they will still need to access different system directories, and we don't know fully which ones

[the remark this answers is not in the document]
adrianchiris: this should be addressed on a ad-hoc basis (if additional mounts needed)
for now we should try to have those tool part of the container image invoking them from the container,

e,g for mstflint related tools we do so today in sriov network config daemon

pkg/registry/suite_test.go
```diff
@@ -0,0 +1,99 @@
...
+		CRDDirectoryPaths:     []string{filepath.Join("..", "..", "config", "crd", "bases")},
+		ErrorIfCRDPathMissing: true,
+
+		// The BinaryAssetsDirectory is only required if you want to run the tests directly
+		// without call the makefile target test. If not informed it will look for the
+		// default path defined in controller-runtime which is /usr/local/kubebuilder/.
+		// Note that you must have the required binaries setup under the bin directory to perform
+		// the tests directly. When we run make test it will be setup and used automatically.
+		BinaryAssetsDirectory: filepath.Join("..", "..", "bin", "k8s",
+			fmt.Sprintf("1.29.0-%s-%s", runtime.GOOS, runtime.GOARCH)),
```
adrianchiris, line 70: nit: you use k8s 1.31 in go.mod 

should you use envtest of 1.31 or 1.30 if envtest has issues with 1.31?

pkg/host/host.go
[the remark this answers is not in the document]
almaslennikov: Now we install mstflint to our docker image, and only mount /sys dir (strace on all the used commands didn't show any additional dirs that we might need to mount) so far. That said, using the centos image as a base now

pkg/registry/deviceregistry.go
```diff
@@ -0,0 +1,175 @@
...
+type DeviceRegistry struct {
+	client.Client
+
+	hostManager host.HostManager
+	namespace   string
+}
+
+// Constructs a unique CR name based on the device's type and serial number
+func getCRName(deviceType string, serialNumber string) string {
+	return strings.ToLower(deviceType + "-" + serialNumber)
```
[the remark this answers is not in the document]
almaslennikov: Let's add this, it adds clarity for the user when listing CRs

pkg/registry/deviceregistry_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/Mellanox/nic-configuration-operator/api/v1alpha1"
+	"github.com/Mellanox/nic-configuration-operator/pkg/consts"
+	"github.com/Mellanox/nic-configuration-operator/pkg/host/mocks"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
```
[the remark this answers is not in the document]
almaslennikov: Changed the approach a little to use controller manager for starting the device discovery, it allows us to test that indexing by node name works as expected. Otherwise left tests as they were before

pkg/registry/deviceregistry.go
```diff
@@ -0,0 +1,175 @@
...
+		// TODO Trigger manual reconcile on the first run, could be the run after reboot
+
+		// Device was processed, cleaning it from the map
+		delete(observedDevices, nicDeviceCR.Status.SerialNumber)
+	}
+
+	// Remaining devices don't have CR representation, need to create a CR
+	for _, deviceStatus := range observedDevices {
+		deviceName := getCRName(deviceStatus.Type, deviceStatus.SerialNumber)
+		device := &v1alpha1.NicDevice{
```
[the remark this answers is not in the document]
almaslennikov: According to [this](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications) we can't set a node as owner for our resources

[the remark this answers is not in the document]
adrianchiris: here is what i see in the link you sent:

> Cross-namespace owner references are disallowed by design. **Namespaced dependents can specify cluster-scoped** or namespaced owners.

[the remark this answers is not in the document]
adrianchiris: if you prefer not to set owner Ref which will delete NicDevice if node deleted. then please add logic to handle the following case:
- Node removed from k8s
- NICDevices associated with that node should be garbage collected.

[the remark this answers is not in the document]
almaslennikov: I see now that there is similar logic in Maintenance operator. I'll verify that it works and if so would prefer the NicDevices to be cleaned up automatically :)

Dockerfile.nic-configuration-daemon
```diff
@@ -20,16 +20,13 @@ COPY ./ ./
...
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
-#RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager cmd/maintenance-manager/main.go
 RUN --mount=type=cache,target=/go/pkg/mod/ GO_GCFLAGS=${GCFLAGS} make build-daemon
 
-# Use distroless as minimal base image to package the nic-configuration-daemon binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-#FROM gcr.io/distroless/static:nonroot
 FROM quay.io/centos/centos:stream9
+ARG MSTFLINT=mstflint
+RUN ARCH_DEP_PKGS=$(if [ "$(uname -m)" != "s390x" ]; then echo -n ${MSTFLINT} ; fi) && yum -y install hwdata $ARCH_DEP_PKGS && yum clean all
```
adrianchiris, line 27: i dont think we care about s390x arch.

internal/controller/devicediscovery_controller.go
```diff
@@ -0,0 +1,172 @@
...
+		log.Log.Error(err, "failed to list NicDevice CRs")
+		return err
+	}
+
+	for _, nicDeviceCR := range list.Items {
+		observedDeviceStatus, exists := observedDevices[nicDeviceCR.Status.SerialNumber]
+
+		if !exists {
+			// Need to delete this CR, it doesn't represent the observedDevice on host anymore
+			err = d.Client.Delete(ctx, &nicDeviceCR)
```
adrianchiris, line 83: One thing im worried about here: 
in DiscoverNicDevices its enough for any command to fail and the device is skipped thus not present in observedDevices which will trigger deletion of NICDevices (loosing their condition, state etc).

perhaps we should be strict in DiscoverNicDevices ? that is:

1. if we fail to list PCI devices -> we fail and retry
2. if we fail to get any information on a mlnx device -> we fail and retry
3. retry in this case should be sooner than 5 min imo.

WDYT ?

adrianchiris: with current logic, device will disappear for at least 5 minutes, then get recreated and reconfigured. (loosing its last applied state)

almaslennikov: One thing to notice is that if we make DiscoverNicDevices stricter, one faulty device will block all devices on that node from creating if we consistently fail to get some info about it, which could be okay as long as we report it properly.

WDYT?

almaslennikov: So I would prefer making DiscoverNicDevices a bit stricter and fail if we can't list PCI devices. I wouldn't fail when we can't get some device information on Mellanox devices. For example, there can be management interfaces like this:
`3b:00.2 DMA controller [0801]: Mellanox Technologies MT42822 BlueField-2 SoC Management Interface [15b3:c2d3]`
That do not support mstvpd calls. So we either make a list of device types that we want to skip (there can be others) or not fail on PN/SN/FW version/PSID retrievals.

When we do fail, though, we should requeue right away

adrianchiris: discovery logic should skip non network PCI device class with IB or ETH subclass see [1]
we do the this in sriov device plugin.

for what is left id expect this to work. anyway i leave it up to you to decide how to handle.

[1] https://pcisig.com/sites/default/files/files/PCI_Code-ID_r_1_11__v24_Jan_2019.pdf

almaslennikov: Makes sense to me. Added filtering based on the PCI device class, and now we fail when some properties of mellanox network device couldn't be retrieved.

almaslennikov:
@adrianchiris addressed all your comments, PTAFL

internal/controller/devicediscovery_controller.go
```diff
@@ -0,0 +1,235 @@
...
+
+		err = d.Client.Create(ctx, device)
+
+		if err := setOwnerRef(ctx, d.Client, node, device); err != nil {
+			log.Log.Error(err, "failed to set owner reference for device", "device", device)
+			continue
+		}
+
+		if apierrors.IsAlreadyExists(err) {
+			// Device already exists but was not matched by SerialNumber, which means the status was not applied properly
```
adrianchiris, line 165: how can this happen ?

almaslennikov: If the device gets created but subsequent status update fails. We can't combine those two in the same API call. It's really a corner case but still can happen

internal/controller/devicediscovery_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	for _, deviceStatus := range observedDevices {
+		deviceName := d.getCRName(deviceStatus.Type, deviceStatus.SerialNumber)
+		device := &v1alpha1.NicDevice{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      deviceName,
+				Namespace: d.namespace,
+			},
+		}
+
+		err = d.Client.Create(ctx, device)
```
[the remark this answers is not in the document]
almaslennikov: Fixed and added a unit test to check owner references

internal/controller/devicediscovery_controller.go
adrianchiris: that means device serial number is empty -> it will get deleted  at L97

up2u if you want to keep this logic, im probably missing something here.

internal/controller/devicediscovery_controller.go
```diff
@@ -0,0 +1,211 @@
...
+	for _, deviceStatus := range observedDevices {
+		deviceName := d.getCRName(deviceStatus.Type, deviceStatus.SerialNumber)
+		device := &v1alpha1.NicDevice{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      deviceName,
+				Namespace: d.namespace,
+			},
+		}
+
+		err := controllerutil.SetOwnerReference(node, device, d.Client.Scheme())
```
adrianchiris, line 132: if this can fail then i would check for it.

adrianchiris (approved):
final round of comments. 

LGTM, once addressed feel free to self-merge.

internal/controller/devicediscovery_controller.go
almaslennikov: Not really. We list devices based on their `node` field in the status. If the status is empty, the cr will not match the selector. Thus, we try to recover a device without status

PR #5, opened 2024-08-30
 Add implementation of NicConfigurationTemplate reconciler

almaslennikov:
Dependent on https://github.com/Mellanox/nic-configuration-operator/pull/4

Only the last commit is relevant to this PR

This reconciler is getting a list of NicDevice matching the selectors and applies the template spec to them


internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -40,26 +46,121 @@ type NicConfigurationTemplateReconciler struct {
...
+	}
+
+	// Create a map to easily look up nodes by name
+	matchingNodes := make(map[string]struct{})
+	for _, node := range nodeList.Items {
+		matchingNodes[node.Name] = struct{}{}
+	}
+
+	matchingNicTypes := make(map[string]struct{})
+	for _, nicType := range strings.Split(template.Spec.NicSelector.NicType, consts.NicTypeSelectorDelimiter) {
```
adrianchiris, line 72: it was probably not clear in the doc, but you can only have a single type specified  no need to support multiple types.
(else we would have made that into a list)

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -18,13 +18,19 @@ package controller
 
 import (
 	"context"
+	"reflect"
+	"slices"
+	"strings"
+	"time"
 
+	v1alpha1 "github.com/Mellanox/nic-configuration-operator/api/v1alpha1"
```
adrianchiris, line 26: nit: fix import order.

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -40,26 +46,121 @@ type NicConfigurationTemplateReconciler struct {
...
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.17.3/pkg/reconcile
+// Reconcile reconciles the NicConfigurationTemplate object
 func (r *NicConfigurationTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	template := &v1alpha1.NicConfigurationTemplate{}
+
+	err := r.Get(ctx, req.NamespacedName, template)
+	if err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
```
adrianchiris, line 55: shouldnt we clear the spec in NicDevices which reference this in case object was not found ?

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -40,26 +46,121 @@ type NicConfigurationTemplateReconciler struct {
...
+	for _, nicType := range strings.Split(template.Spec.NicSelector.NicType, consts.NicTypeSelectorDelimiter) {
+		matchingNicTypes[nicType] = struct{}{}
+	}
+
+	// Get the list of NicDevices matching the NicSelectorSpec
+	deviceList := &v1alpha1.NicDeviceList{}
+	err = r.List(ctx, deviceList)
+	if err != nil {
+		log.Log.Error(err, "Failed to list NicDevices")
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, err
```
adrianchiris, line 81: you return error , so you can keep ctrl.Result{}  as it has no meaning in such case.

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -40,26 +46,121 @@ type NicConfigurationTemplateReconciler struct {
...
+			err = r.Update(ctx, matchingDevice)
+			if err != nil {
+				lastErr = err
+				log.Log.Error(err, "Failed to update NicDevice spec", "device", matchingDevice.Name)
+			}
+		}
+	}
+
+	if lastErr != nil {
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, lastErr
```
adrianchiris, line 119: you return error , so you can keep ctrl.Result{} as it has no meaning in such case.

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -40,26 +46,121 @@ type NicConfigurationTemplateReconciler struct {
...
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, err
+	}
+
+	matchingDevices := []*v1alpha1.NicDevice{}
+	for _, device := range deviceList.Items {
+		if nicDeviceMatchesSelectors(&device, matchingNodes, matchingNicTypes, template.Spec.NicSelector.PciAddresses, template.Spec.NicSelector.SerialNumbers) {
+			matchingDevices = append(matchingDevices, &device)
+		}
+	}
+
```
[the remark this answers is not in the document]
almaslennikov: Yes, thanks!

adrianchiris:
@almaslennikov  can you rebase this one ?

almaslennikov:
> @almaslennikov can you rebase this one ?

done

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -43,26 +57,254 @@ type NicConfigurationTemplateReconciler struct {
...
-	_ = log.FromContext(ctx)
+	if req.Name != nicConfigurationTemplateSyncEventName || req.Namespace != "" {
+		return reconcile.Result{}, nil
+	}
+	log.Log.Info("Reconciling NicConfigurationTemplates")
+
+	templateList := &v1alpha1.NicConfigurationTemplateList{}
+	err := r.List(ctx, templateList)
+	if err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
```
adrianchiris, line 70: im not sure list items can fail with not found. even if there are no templates you will get a list object with empty items.

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -43,26 +57,254 @@ type NicConfigurationTemplateReconciler struct {
...
+		log.Log.Error(err, "Failed to list NicDevices")
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+	log.Log.V(2).Info(fmt.Sprintf("listed %d devices: %s", len(deviceList.Items), deviceList.Items))
+
+	nodeList := &v1.NodeList{}
+	err = r.List(ctx, nodeList)
+	if err != nil {
+		log.Log.Error(err, "Failed to list cluster nodes")
+		return ctrl.Result{}, client.IgnoreNotFound(err)
```
adrianchiris, line 86: why do we need client.IgnoreNotFound(err) ?

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -43,26 +57,254 @@ type NicConfigurationTemplateReconciler struct {
...
+		if !ok {
+			log.Log.Info("device doesn't match any node, skipping", "device", device.Name)
+			continue
+		}
+
+		var matchingTemplates []*v1alpha1.NicConfigurationTemplate
+
+		for _, template := range templateList.Items {
+			if !deviceMatchesSelectors(&device, &template, node) {
+				err := r.dropDeviceFromStatus(ctx, device.Name, &template)
```
adrianchiris, line 106: lets have a single API call to update status instead of many even if it means we do more work here.

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -43,26 +57,254 @@ type NicConfigurationTemplateReconciler struct {
...
+	return nil
+}
+
+func (r *NicConfigurationTemplateReconciler) applyTemplateToDevice(ctx context.Context, device *v1alpha1.NicDevice, template *v1alpha1.NicConfigurationTemplate) error {
+	log.Log.V(2).Info(fmt.Sprintf("Applying template %s to device %s", template.Name, device.Name))
+
+	updateSpec := false
+	if device.Spec.Configuration == nil {
+		updateSpec = true
+		device.Spec.Configuration = &v1alpha1.NicDeviceConfigurationSpec{}
```
adrianchiris, line 181: do we need to continue to the other if conditions ?

adrianchiris (commented):
please add controller test to cover the added logic.

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -43,26 +57,254 @@ type NicConfigurationTemplateReconciler struct {
...
+	nicDeviceEventHandler := handler.Funcs{
+		UpdateFunc: func(ctx context.Context, e event.UpdateEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+			log.Log.Info("Enqueuing sync for update event", "resource", e.ObjectNew.GetName())
+			qHandler(q)
+		},
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&configurationnetv1alpha1.NicConfigurationTemplate{}).
+		For(&v1alpha1.NicConfigurationTemplate{}).
```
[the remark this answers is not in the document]
almaslennikov: I don't think so, no

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -43,26 +57,254 @@ type NicConfigurationTemplateReconciler struct {
...
+	if err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
+	log.Log.V(2).Info(fmt.Sprintf("listed %d templates: %s", len(templateList.Items), templateList.Items))
+
+	deviceList := &v1alpha1.NicDeviceList{}
+	err = r.List(ctx, deviceList)
+	if err != nil {
+		log.Log.Error(err, "Failed to list NicDevices")
+		return ctrl.Result{}, client.IgnoreNotFound(err)
```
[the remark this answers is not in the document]
almaslennikov: You're right, we don't for the list :)

internal/controller/nicconfigurationtemplate_controller.go
almaslennikov: Yes we do because NicDevice and NicConfigurationTemplate store different types in the root spec object and we have to compare / apply template field by field

internal/controller/nicconfigurationtemplate_controller.go
almaslennikov: Done, now status updates once for each template

almaslennikov:
> please add controller test to cover the added logic.

By a freaky accident with rebasing, the utests disappeared. git reflog to the rescue, however, and now they're back!

almaslennikov:
Closed by mistake, reopened #12

PR #6, opened 2024-09-15
Add Nic Device Controller

almaslennikov:
- The controller follows the attached diagram
- Configuration of the device is done via HostManager
- Long commands have cancellable context


almaslennikov:
Closed by mistake, reopened #13

PR #12, opened 2024-09-18
Add implementation of NicConfigurationTemplate reconciler

almaslennikov:
This reconciler is getting a list of NicDevice matching the selectors and applies the template spec to them

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,5 @@
-# VERSION defines the project version for the bundle.
-# Update this value when you upgrade the version of your project.
-# To re-generate a bundle for another specific version without changing the standard setup, you can:
-# - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
-# - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.1
+# Version information
+include Makefile.version
 
 # CHANNELS define the bundle channels used in the bundle.
 # Add a new line here if you would like to change its default config. (E.g CHANNELS = "candidate,fast,stable")
@@ -68,6 +64,13 @@ endif
 # scaffolded by default. However, you might want to replace it to use other
 # tools. (i.e. podman)
 CONTAINER_TOOL ?= docker
+
+# Build args
+TARGETOS ?= $(shell go env GOOS)
+TARGETARCH ?= $(shell go env GOARCH)
+GO_BUILD_OPTS ?= CGO_ENABLED=0 GOOS=$(TARGETOS) GOARCH=$(TARGETARCH)
+GO_LDFLAGS ?= $(VERSION_LDFLAGS)
+GO_GCFLAGS ?=
 
 # Setting SHELL to bash allows bash commands to be executed by recipes.
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
@@ -137,7 +140,7 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
 build: manifests generate fmt vet build-manager build-daemon
 
 build-manager: ## Build manager binary.
-	go build -o build/manager cmd/manager/main.go
+	$(GO_BUILD_OPTS) go build -ldflags $(GO_LDFLAGS) -gcflags="$(GO_GCFLAGS)" -o build/manager cmd/manager/main.go
 
 build-daemon: ## Build nic-configuration-daemon binary.
 	go build -o build/nic-configuration-daemon cmd/nic-configuration-daemon/main.go
@@ -165,7 +168,7 @@ docker-push: ## Push docker image with the manager.
 # - have enabled BuildKit. More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 # - be able to push the image to your registry (i.e. if you do not set a valid value via OPERATOR_IMAGE_TAG=<myregistry/image:<tag>> then the export will fail)
 # To adequately provide solutions that are compatible with multiple platforms, you should consider using this option.
-PLATFORMS ?= linux/arm64,linux/amd64,linux/s390x,linux/ppc64le
+PLATFORMS ?= linux/arm64,linux/amd64
 .PHONY: docker-buildx
 docker-buildx: ## Build and push docker image for the manager for cross-platform support
 	# copy existing Dockerfile and insert --platform=${BUILDPLATFORM} into Dockerfile.cross, and preserve the original Dockerfile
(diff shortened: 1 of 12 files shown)


internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -43,26 +57,262 @@ type NicConfigurationTemplateReconciler struct {
...
+		if err != nil {
+			log.Log.Error(err, "failed to update template status", "template", template.Name)
+			return ctrl.Result{}, err
+		}
+	}
 
 	return ctrl.Result{}, nil
 }
 
+func (r *NicConfigurationTemplateReconciler) dropDeviceFromStatus(ctx context.Context, deviceName string, template *v1alpha1.NicConfigurationTemplate) error {
```
ykulazhenkov, line 174: error is never used, let's drop it

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -43,26 +57,262 @@ type NicConfigurationTemplateReconciler struct {
...
+		if !ok {
+			log.Log.Info("device doesn't match any node, skipping", "device", device.Name)
+			continue
+		}
+
+		var matchingTemplates []*v1alpha1.NicConfigurationTemplate
+
+		for _, template := range templates {
+			if !deviceMatchesSelectors(&device, template, node) {
+				err := r.dropDeviceFromStatus(ctx, device.Name, template)
```
ykulazhenkov, line 111: this function never returns errors

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -43,26 +57,262 @@ type NicConfigurationTemplateReconciler struct {
...
 
-	// TODO(user): your logic here
+			templateNames := []string{}
+			for _, template := range matchingTemplates {
+				templateNames = append(templateNames, template.Name)
+			}
+			joinedTemplateNames := strings.Join(templateNames, ",")
+			err = fmt.Errorf("device matches several configuration templates: %s, %s", device.Name, joinedTemplateNames)
+			log.Log.Error(err, "Device matches several configuration templates, deleting its spec and emitting an error event", "device", device.Name)
+			return ctrl.Result{}, r.handleErrorSeveralMatchingTemplates(ctx, &device, joinedTemplateNames)
```
adrianchiris, line 146: why do we want to exit early and not continue to other devices ?
i would return only if error happened (then we reconcile again)

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -43,26 +57,262 @@ type NicConfigurationTemplateReconciler struct {
...
+		if len(matchingTemplates) == 0 {
+			log.Log.V(2).Info("Device doesn't match any configuration template, resetting the spec", "device", device.Name)
+			device.Spec.Configuration = nil
+			err = r.Update(ctx, &device)
+			if err != nil {
+				log.Log.Error(err, "Failed to update device's spec", "device", device)
+				return ctrl.Result{}, err
+			}
+			continue
+		} else if len(matchingTemplates) > 1 {
```
ykulazhenkov, line 131: You can move this `else if` condition to an individual `if` case

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -43,26 +57,262 @@ type NicConfigurationTemplateReconciler struct {
...
+			for _, template := range matchingTemplates {
+				err := r.dropDeviceFromStatus(ctx, device.Name, template)
+				if err != nil {
```
ykulazhenkov, line 134: the function never return errors

internal/controller/nicconfigurationtemplate_controller.go
ykulazhenkov, line 146: I think in this case the device name will never be removed from the template status. You removed it from the struct, but didn't updated the API.

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -43,26 +57,262 @@ type NicConfigurationTemplateReconciler struct {
...
+			log.Log.Error(err, "failed to update template status", "template", template.Name)
+			return ctrl.Result{}, err
+		}
+	}
 
 	return ctrl.Result{}, nil
 }
 
+func (r *NicConfigurationTemplateReconciler) dropDeviceFromStatus(ctx context.Context, deviceName string, template *v1alpha1.NicConfigurationTemplate) error {
+	if slices.Contains(template.Status.NicDevices, deviceName) {
```
ykulazhenkov, line 175: nit: you can save call of Contains - Index will return -1 if the element is not found

internal/controller/nicconfigurationtemplate_controller_test.go
```diff
@@ -0,0 +1,386 @@
...
+		Expect(k8sClient.Create(ctx, device6)).To(Succeed())
+		device6.Status = v1alpha1.NicDeviceStatus{
+			Node:         validNode.Name,
+			Type:         "ConnectX6",
+			Ports:        []v1alpha1.NicDevicePortSpec{{PCI: "0000:3b:00.0"}},
+			SerialNumber: "serialNumber3",
+		}
+		Expect(k8sClient.Status().Update(ctx, device6)).To(Succeed())
+
+		Expect(k8sClient).To(Not(BeNil()))
```
ykulazhenkov, line 225: nit: it looks like it is a bit too late to check this here :) You already used the client multiple times

internal/controller/nicconfigurationtemplate_controller.go
almaslennikov: Changed this one to return an error only if the function failed. Templates are updated together at the end of the reconcile loop

internal/controller/nicconfigurationtemplate_controller_test.go
almaslennikov: You're right :) Dropped this check, it's not needed actually

internal/controller/nicconfigurationtemplate_controller.go
almaslennikov: Thanks for the tip! Applied it

internal/controller/nicconfigurationtemplate_controller.go
almaslennikov: Yep, thanks, dropped the error handling

Makefile.version
```diff
@@ -0,0 +1,39 @@
...
+ifneq (,$(GIT_TAG))
+    RELEASE_STATUS = ",released"
+endif
+
+ifeq (,$(VERSION))
+    VERSION = $(GIT_TAG_LAST)
+endif
+
+# Add version/commit/date to linker flags
+VERSION_LDFLAGS = "-X github.com/Mellanox/nic-configuration-operator/internal/version.version=${VERSION} \
```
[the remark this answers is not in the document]
almaslennikov: Those aren't files but linker flags to build metadata into the binary

Makefile
```diff
@@ -1,9 +1,12 @@
+# Version information
+include Makefile.version
+
 # VERSION defines the project version for the bundle.
 # Update this value when you upgrade the version of your project.
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.1
+# VERSION ?= 0.0.1
```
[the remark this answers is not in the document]
almaslennikov: Dropped it, thanks

internal/controller/nicconfigurationtemplate_controller.go
almaslennikov: You're right, not it only returns if an error occurred

internal/controller/nicconfigurationtemplate_controller.go
```diff
@@ -43,26 +57,262 @@ type NicConfigurationTemplateReconciler struct {
...
+	nicDeviceEventHandler := handler.Funcs{
+		UpdateFunc: func(ctx context.Context, e event.UpdateEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+			log.Log.Info("Enqueuing sync for update event", "resource", e.ObjectNew.GetName())
+			qHandler(q)
+		},
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&configurationnetv1alpha1.NicConfigurationTemplate{}).
+		For(&v1alpha1.NicConfigurationTemplate{}).
```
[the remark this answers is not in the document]
almaslennikov: Dropped it, thx

Makefile.version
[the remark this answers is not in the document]
ykulazhenkov: Yes, but usually these flags are used to set values in existing files, like this https://github.com/Mellanox/maintenance-operator/blob/main/internal/version/version.go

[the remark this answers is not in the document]
almaslennikov: yeah, you're right. Added version handling in the manager binary

adrianchiris (approved):

rollandf (approved):

PR #13, opened 2024-09-18
Add Nic Device Controller

almaslennikov:
The controller follows the attached diagram
Configuration of the device is done via HostManager
Long commands have cancellable context


internal/controller/nicdevice_controller.go
```diff
@@ -0,0 +1,550 @@
...
+
+type nicDeviceConfigurationStatus struct {
+	device                 *v1alpha1.NicDevice
+	nvConfigUpdateRequired bool
+	rebootRequired         bool
+	lastStageError         error
+}
+
+// Reconcile reconciles the NicConfigurationTemplate object
+func (r *NicDeviceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
tobiasgiese, line 71: I think it makes sense to put the reconciliation in multiple helper methods. Otherwise, it can pretty hard to read and maintain it. WDYT?

internal/controller/nicdevice_controller.go
```diff
@@ -0,0 +1,550 @@
...
+	if !configStatuses.nvConfigReadyForAll() {
+		return ctrl.Result{Requeue: true, RequeueAfter: requeueTime}, nil
+	}
+
+	err = r.applyRuntimeConfig(ctx, configStatuses)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if configStatuses.rebootRequired() {
```
tobiasgiese, line 165: It feels a bit weird to have the same `if reboot required` code in the same method. Maybe we can optimize this? But I don't have a better idea but "simply requeue" and ujse the "if reboot required" from above

internal/controller/nicdevice_controller.go
```diff
@@ -0,0 +1,550 @@
...
+	}
+
+	return nil
+}
+
+// handleReboot schedules maintenance and reboots the node if maintenance is allowed
+// Before rebooting the node, strips LastAppliedState annotations from all devices
+// returns true if requeue of the reconcile request is required, false otherwise
+// return err if encountered an error while performing maintenance scheduling / reboot
+func (r *NicDeviceReconciler) handleReboot(ctx context.Context, statuses nicDeviceConfigurationStatuses) (bool, error) {
```
tobiasgiese, line 272: I think it's better to return with a controller result (i.e., `ctrl.Result{RequeueAfter: requeueTime`) instead of a boolean. The boolean makes it harder to understand what's happening. `handleReboot` returns with `true` which usually indicates the the reboot has been handled

pkg/host/configvalidation.go
```diff
@@ -0,0 +1,322 @@
...
+		if len(ports) == 2 {
+			actualMaxReadReqSize, err = v.utils.GetMaxReadRequestSize(ports[1].PCI)
+			if err != nil {
+				log.Log.Error(err, "can't validate maxReadReqSize", "device", device.Name)
+				return false, err
+			}
+			if actualMaxReadReqSize != desiredMaxReadReqSize {
+				return false, nil
+			}
+		}
```
tobiasgiese, line 259: why don't simply looping over `ports`?

pkg/host/configvalidation.go
```diff
@@ -0,0 +1,322 @@
...
+			trust = template.RoceOptimized.Qos.Trust
+			pfc = template.RoceOptimized.Qos.PFC
+		} else {
+			trust = "dscp"
+			pfc = "0,0,0,1,0,0,0,0"
+		}
+	} else {
+		trust = "pcp"
+		pfc = "0,0,0,0,0,0,0,0"
+	}
```
tobiasgiese, line 315: nit: You could set this as the default values while initializing the vars and remove the else block.

pkg/host/host.go
```diff
@@ -130,28 +135,195 @@ func (h hostManager) DiscoverNicDevices() (map[string]v1alpha1.NicDeviceStatus,
...
+			log.Log.Error(err, "failed to apply runtime configuration", "device", device)
+			return err
+		}
+		if len(ports) == 2 {
+			err = h.hostUtils.SetMaxReadRequestSize(ports[1].PCI, desiredMaxReadReqSize)
+			if err != nil {
+				log.Log.Error(err, "failed to apply runtime configuration", "device", device)
+				return err
+			}
+		}
```
tobiasgiese, line 308: same here, a loop would be sufficcient. Also for the next one

pkg/host/utils.go
```diff
@@ -141,6 +166,123 @@ func (h *hostUtils) GetFirmwareVersionAndPSID(pciAddr string) (string, string, e
...
+
+	// Parse the output for LnkSta
+	scanner := bufio.NewScanner(strings.NewReader(string(output)))
+
+	for scanner.Scan() {
+		line := strings.TrimSpace(strings.ToLower(scanner.Text()))
+
+		if strings.HasPrefix(line, consts.LinkStatsPrefix) {
+			match := linkSpeedRegexp.FindStringSubmatch(line)
+			if len(match) == 2 {
```
tobiasgiese, line 191: nit: to stick forther to the left side

```go
if len(match) != 2 {
    continue
}
```

pkg/host/utils.go
```diff
@@ -141,6 +166,123 @@ func (h *hostUtils) GetFirmwareVersionAndPSID(pciAddr string) (string, string, e
...
+
+	linkSpeedRegexp := regexp.MustCompile(`speed\s+([0-9.]+)gt/s`)
+
+	// Parse the output for LnkSta
+	scanner := bufio.NewScanner(strings.NewReader(string(output)))
+
+	for scanner.Scan() {
+		line := strings.TrimSpace(strings.ToLower(scanner.Text()))
+
+		if strings.HasPrefix(line, consts.LinkStatsPrefix) {
```
tobiasgiese, line 189: same here: continue if it has not the prefix.

tobiasgiese (commented):
This was just a high-level review, will do another round later

Makefile
```diff
@@ -175,6 +175,13 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 	- $(CONTAINER_TOOL) buildx build --push --platform=$(PLATFORMS) --tag ${OPERATOR_IMAGE_TAG} -f Dockerfile.cross .
 	- $(CONTAINER_TOOL) buildx rm project-v3-builder
 	rm Dockerfile.cross
+	# same for config-daemon
+	sed -e '1 s/\(^FROM\)/FROM --platform=\$$\{BUILDPLATFORM\}/; t' -e ' 1,// s//FROM --platform=\$$\{BUILDPLATFORM\}/' Dockerfile.nic-configuration-daemon > Dockerfile.nic-configuration-daemon.cross
```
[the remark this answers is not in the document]
almaslennikov: Because we don't necessarily want to build cross-platform (make docker-build target). Essentially, it's the code, auto-generated by operator-sdk, slightly changed to also build the daemon

cmd/nic-configuration-daemon/main.go
```diff
@@ -60,13 +61,21 @@ func main() {
...
 	hostManager := host.NewHostManager(hostUtils)
+	maintenanceManager := maintenance.New()
 
 	deviceDiscovery := controller.NewDeviceRegistry(mgr.GetClient(), hostManager, nodeName, namespace)
 	if err = mgr.Add(deviceDiscovery); err != nil {
 		log.Log.Error(err, "unable to add device discovery runnable")
 		os.Exit(1)
 	}
 
+	nicDeviceReconciler := controller.NewNicDeviceReconciler(mgr.GetClient(), mgr.GetScheme(), nodeName, namespace, hostManager, maintenanceManager)
```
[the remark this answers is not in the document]
tobiasgiese:
```go
	nicDeviceReconciler := controller.NewNicDeviceReconciler(
		Client: mgr.GetClient(),
		Scheme: mgr.GetScheme(),
		[...]
	)
```

We are doing it like this in our other controllers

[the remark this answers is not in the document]
almaslennikov: Changed to your suggestion

internal/controller/nicdevice_controller.go
almaslennikov: Broken down the reconcile func a bit, hope you like it now :)

internal/controller/nicdevice_controller.go
```diff
@@ -0,0 +1,550 @@
...
+	selectorFields := fields.OneTermEqualSelector("status.node", r.nodeName)
+
+	err := r.Client.List(ctx, devices, &client.ListOptions{FieldSelector: selectorFields})
+	if err != nil {
+		log.Log.Error(err, "failed to list NicDevice CRs")
+		return ctrl.Result{}, err
+	}
+
+	if len(devices.Items) == 0 {
+		err = r.maintenanceManager.ReleaseMaintenance(ctx)
```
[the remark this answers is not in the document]
almaslennikov: We own the maitenance object and it won't release the maintenance from the cluster if there are other maintenance requests created on the same node by someone else

internal/controller/nicdevice_controller.go
almaslennikov: Those are actually different reboots (after nv config apply and runtime config apply) and we don't keep the state in between the calls

tobiasgiese: But this means also that we can be stuck in a reboot loop, right?
If we do the nv config apply, reboot and the runtime config apply fails every time it will nv config apply again, reboot and runtime config apply fails... again. Is this an possible edge case?

almaslennikov: It works like this:
1. ValidateNvConfig returns whether nvConfig change or reboot are required
2. If nv config change is requested it will apply the new config
3. if the reboot was requested (either before or after nv config change) it will reboot the node
4. The second time the nv config is already applied so the reboot will not be requested from this step.
5. If the runtime config failed to apply it will retry again. Reboot is only required if we've had some previously setup configuration that does not match the spec (we save that as lastAppliedState annotation in the CR)

There potentially is a reboot loop when nv config was not properly applied after the reboot and we are looking into whether it's even possible on a healthy device

internal/controller/nicdevice_controller.go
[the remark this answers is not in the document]
almaslennikov: Yeah, so it checks if we own some maintenance already and if so, we release it

rollandf (approved):

cmd/nic-configuration-daemon/main.go
```diff
@@ -31,6 +33,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	err = maintenanceoperator.AddToScheme(scheme)
+	if err != nil {
+		log.Log.Error(err, "failed to load maintenance operator to scheme")
+		os.Exit(1)
+	}
```
tobiasgiese, line 40: nit:

```suggestion
	utilruntime.Must(maintenanceoperator.AddToScheme(scheme))
```

with `import utilruntime "k8s.io/apimachinery/pkg/util/runtime"`

cmd/nic-configuration-daemon/main.go
```diff
@@ -59,14 +67,29 @@ func main() {
 	}
 
 	hostUtils := host.NewHostUtils()
-	hostManager := host.NewHostManager(hostUtils)
+	hostManager := host.NewHostManager(nodeName, hostUtils)
+	maintenanceManager := maintenance.New(mgr.GetClient(), hostUtils, nodeName, namespace)
```
tobiasgiese, line 71: We should follow the same pattern as in L79

pkg/host/configvalidation.go
```diff
@@ -0,0 +1,312 @@
...
+	secondPortPresent := len(device.Status.Ports) > 1
+	pciAddr := device.Status.Ports[0].PCI
+
+	if template.NumVfs == 0 {
+		desiredParameters[consts.SriovEnabledParam] = consts.NvParamFalse
+		desiredParameters[consts.SriovNumOfVfsParam] = "0"
+	} else {
+		desiredParameters[consts.SriovEnabledParam] = consts.NvParamTrue
+		desiredParameters[consts.SriovNumOfVfsParam] = strconv.Itoa(template.NumVfs)
+	}
```
tobiasgiese, line 92: nit: to follow the go-style

```suggestion
	desiredParameters[consts.SriovEnabledParam] = consts.NvParamFalse
	desiredParameters[consts.SriovNumOfVfsParam] = "0"
	if template.NumVfs > 0 {
		desiredParameters[consts.SriovEnabledParam] = consts.NvParamTrue
		desiredParameters[consts.SriovNumOfVfsParam] = strconv.Itoa(template.NumVfs)
	}
```

tobiasgiese (approved):

cmd/nic-configuration-daemon/main.go
almaslennikov: I would rather not because that way we hide the implementation inside the package

PR #16, opened 2024-09-22
feat: Implement firmware version check

e0ne:


internal/controller/devicediscovery_controller.go
```diff
@@ -63,6 +65,28 @@ func setInitialsConditionsForDevice(device *v1alpha1.NicDevice) {
 	meta.SetStatusCondition(&device.Status.Conditions, condition)
 }
 
+func setFwConfigConditionsForDevice(device *v1alpha1.NicDevice) {
+	currentFirmware := helper.NicFirmwareMap[device.Status.FirmwareVersion]
+	recommendedFirmware := helper.NicFirmwareMap[device.Status.Type]
+	if currentFirmware == recommendedFirmware {
```
e0ne, line 71: @adrianchiris do we need to have '>=' here?

adrianchiris: Hi @e0ne following our discussion,

we should try to get mofed version from the system then check against recommended firmware version for the specific mofed version.

how we get the set of firmware versions recommended for  a mofed version should be queriable from mofed scripts or config map if we have no choice.

e0ne: until mofed scripts implement this feature, let's use configmap

pkg/helper/helper.go
```diff
@@ -0,0 +1,57 @@
...
+		return err
+	}
+	for _, v := range cm.Data {
+		NicFirmwareMap = append(NicFirmwareMap, v)
+	}
+
+	return nil
+}
+
+func GetFwVersion(deviceId, ofed string) string {
```
ykulazhenkov, line 49: Do we want to add some sanity checks here? What will happen if the ConfigMap has invalid format?

e0ne: since configmap is deployed by the operator, it will be updated with a correct format during the next reconcile loop, so I don't think we need to handle a case when user updates it manually with an incorrect data

e0ne: it makes sense, I'll update my pr

deployment/nic-configuration-operator-chart/templates/supported-nic-firmware-configmap.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: supported-nic-firmware
+data:
+  Nvidia_mlx5_ConnectX-4: "1013 24.07-0.6.1 12.28.2006"
+  Nvidia_mlx5_ConnectX-5: "1017 24.07-0.6.1 16.35.4030"
+  Nvidia_mlx5_ConnectX-5_Ex: "1019 24.07-0.6.1 16.35.4030"
+  Nvidia_mlx5_ConnectX-6: "101b 24.07-0.6.1 20.42.1000"
+  Nvidia_mlx5_ConnectX-6_Dx: "101d 24.07-0.6.1 22.42.1000"
+  Nvidia_mlx5_ConnectX-6_Lx: "101f 24.07-0.6.1 26.42.1000"
+  Nvidia_mlx5_ConnectX-7: "1021 24.07-0.6.1 28.42.1000"
+  Nvidia_mlx5_MT42822_BlueField-2_integrated_ConnectX-6_Dx: "a2d6 24.07-0.6.1 24.42.1000"
```
almaslennikov, line 13: Do we also want to deploy this cfg map with make deploy?

pkg/helper/helper.go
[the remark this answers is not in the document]
e0ne: done

deployment/nic-configuration-operator-chart/templates/supported-nic-firmware-configmap.yaml
e0ne: we do:)

almaslennikov (approved):

ykulazhenkov (approved):
LGTM

PR #18, opened 2024-10-01
Add cfgmap permissions to RBAC

almaslennikov:


config/rbac/role.yaml
```diff
@@ -4,6 +4,13 @@ kind: ClusterRole
 metadata:
   name: manager-role
 rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - create
```
e0ne, line 12: do we really need 'create' permissions?

e0ne (approved):
LGTM

config/rbac/role.yaml
almaslennikov: You're right, we don't. Dropped them, left only the 'get' permissions

PR #19, opened 2024-10-01
fix: Validate the linkType correctly when it can and cannot be changed

almaslennikov:
If LINK_TYPE_P1 parameter is not present in the mstconfig defaults, we can't change the link type on the device and should return an error if the actual link type differs from the template. Otherwise, we should not raise an error.

### DIFF
--- a/pkg/host/configvalidation.go
+++ b/pkg/host/configvalidation.go
@@ -89,24 +89,26 @@ func (v *configValidationImpl) ConstructNvParamMapFromTemplate(
 		desiredParameters[consts.SriovNumOfVfsParam] = strconv.Itoa(template.NumVfs)
 	}
 
-	if template.LinkType != "" {
-		// Link type change is not allowed on some devices
-		if _, found := defaultValues[consts.LinkTypeP1Param]; found {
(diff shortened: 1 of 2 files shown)


pkg/host/configvalidation.go
```diff
@@ -89,24 +89,26 @@ func (v *configValidationImpl) ConstructNvParamMapFromTemplate(
...
-	if template.LinkType != "" {
-		// Link type change is not allowed on some devices
-		if _, found := defaultValues[consts.LinkTypeP1Param]; found {
-			linkType := nvParamLinkTypeFromName(string(template.LinkType))
-			desiredParameters[consts.LinkTypeP1Param] = linkType
-			if secondPortPresent {
-				desiredParameters[consts.LinkTypeP2Param] = linkType
-			}
+	// Link type change is not allowed on some devices
+	_, linkTypeCanChange := defaultValues[consts.LinkTypeP1Param]
```
e0ne, line 93: nit: linkTypeCanChange => canChangeLinkType

almaslennikov: Fixed

e0ne (approved):

ykulazhenkov (approved):

PR #23, opened 2024-10-09
fix: allow to enable log levels from helm chart and make deploy target

almaslennikov:


almaslennikov:
Created an issue to move to zap log level literals in the future: https://github.com/Mellanox/nic-configuration-operator/issues/24

e0ne (approved):

PR #27, opened 2024-10-15
Restore setting RoCE configuration

almaslennikov:
The issue breaking RoCE config was in the previous call to setpci that set an incorrect value subsequently breaking the runtime config for the device until the next reboot


ykulazhenkov (approved):

deployment/nic-configuration-operator-chart/templates/config-daemon.yaml
```diff
@@ -24,6 +24,7 @@ spec:
       nodeSelector: {{- toYaml .Values.operator.nodeSelector | nindent 8 }}
       serviceAccountName: {{ include "nic-configuration-operator.serviceAccountName" . }}
       terminationGracePeriodSeconds: 10
+      hostNetwork: true
```
e0ne, line 27: why do we need host network?

almaslennikov: We use netlink socket to apply RoCE settings and we need the host network for that

e0ne (approved):

PR #30, opened 2024-10-24
fix: disable RoCE-specific settings for IB devices

almaslennikov:
This commit removes the configuration of RoCE-specific parameters when the device is operating in InfiniBand (IB) mode. Those settings are intended for RoCE (RDMA over Converged Ethernet) and are irrelevant for IB devices.


ykulazhenkov (approved):
LGTM

e0ne (approved):

PR #33, opened 2024-10-28
small fixes in documentation

almaslennikov:


README.md
```diff
@@ -97,9 +97,11 @@ spec:
 #### Configuration details
 
 * `numVFs`: if provided, configure SR-IOV VFs via nvconfig.
+  * This is a mandatory parameter.
```
tobiasgiese, line 100: We should rather make this field required in our API instead of only documenting this

almaslennikov: It is required, the additional comment here in the docs is for transparency

tobiasgiese: But it is not defined as required via CRD validation. Is this on purpose?

https://github.com/Mellanox/nic-configuration-operator/blob/80db27ef79f1b514091d118834746f6c0391e45b/api/v1alpha1/nicconfigurationtemplate_types.go#L82-L83

xref https://book.kubebuilder.io/reference/markers/crd-validation

almaslennikov: It doesn't have the `omitempty` attribute which doesn't allow to create a template without it.

tobiasgiese: But it is not a strict contract to the API. Additionally it is not user visible while reading the API docs or running `kubectl explain`. With the annotation on it the user will see it automatically

almaslennikov: Okay, let's add it :)

Nevertheless, our CRDs already mark those fields as required: https://github.com/Mellanox/nic-configuration-operator/blob/main/config/crd/bases/configuration.net.nvidia.com_nicconfigurationtemplates.yaml#L169

tobiasgiese (approved):

PR #34, opened 2024-10-29
fix: do not apply QoS settings if RoCE optimizations are disabled + better error handling

almaslennikov:
If RoCE optimizations are disabled in the template, do nothing. If a port doesn't have a network interface and RoCE is requested, report an error.

### DIFF
--- a/pkg/host/configvalidation.go
+++ b/pkg/host/configvalidation.go
@@ -267,10 +267,20 @@ func (v *configValidationImpl) RuntimeConfigApplied(device *v1alpha1.NicDevice) 
 		}
 	}
 
+	// Don't validate QoS settings if neither trust nor pfc changes are requested
+	if desiredTrust == "" && desiredPfc == "" {
+		return true, nil
+	}
+
 	for _, port := range ports {
+		if port.NetworkInterface == "" {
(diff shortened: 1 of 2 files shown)


pkg/host/configvalidation.go
```diff
@@ -267,10 +267,21 @@ func (v *configValidationImpl) RuntimeConfigApplied(device *v1alpha1.NicDevice)
 		}
 	}
 
+	qosConfigurationRequested := desiredTrust != "" || desiredPfc != ""
+
+	if !qosConfigurationRequested {
```
tobiasgiese, line 272: I don't know if this makes it even more complicated understanding the code.
Isn't it easier to understand to say "if both are empty" instead of negating things?
```suggestion
	// Return early if no QoS configuration is requested.
	if desiredTrust == "" && desiredPfc == "" {
```

tobiasgiese (approved):
only one code style nit :) rest lgtm

pkg/host/configvalidation.go
almaslennikov: The user might provide custom values for trust and pfc and leave one empty

ykulazhenkov (approved):

PR #43, opened 2024-11-07
fix: validate if nv config update applied after reboot

almaslennikov:
There might be a case where FW config didn't apply after a reboot because of some error in FW. In this case
we don't want the node to be kept in a reboot loop (FW configured -> reboot -> Config was not applied -> FW configured -> etc.).
To break the reboot loop, we should compare the last time the status was changed to PendingReboot to the node's uptime.
If the node started after the status was changed, we assume the node was rebooted and the config couldn't apply.
In this case, we indicate the error to the user with the status change and emit an error event.

### DIFF
--- a/pkg/host/utils.go
+++ b/pkg/host/utils.go
@@ -25,6 +25,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/Mellanox/rdmamap"
 	"github.com/jaypipes/ghw"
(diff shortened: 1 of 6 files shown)


rollandf (approved):

pkg/host/utils.go
```diff
@@ -630,6 +633,20 @@ func (h *hostUtils) GetOfedVersion() string {
 	return version
 }
 
+// GetHostUptimeSeconds returns the host uptime in seconds
+func (h *hostUtils) GetHostUptimeSeconds() (time.Duration, error) {
+	log.Log.V(2).Info("HostUtils.GetHostUptimeSeconds()")
+	output, err := os.ReadFile("/proc/uptime")
+	if err != nil {
+		log.Log.Error(err, "GetOfedVersion(): failed to read mlx5_core version file")
```
ykulazhenkov, line 641: the error message is wrong

ykulazhenkov (approved):
LGTM

PR #50, opened 2024-11-13
Fix supported-nic-firmware in a helm chart

e0ne:
This is a follow-up PR for #49


almaslennikov (approved):

e0ne:
it's a part of #53 now

PR #61, opened 2025-02-04
Revert "Deploy configuration daemon DS during config template reconcile"

almaslennikov:
Since we've decided not to support a separate OCP bundle for NIC op, let's drop the elaborate daemon deployment scheme and keep deploying both the operator and the daemon set together from the helm chart

This reverts commit cedb2e3af38bcf8ad1aea8776a50f9aada11ae93.

Original PR: https://github.com/Mellanox/nic-configuration-operator/pull/55


almaslennikov:
@adrianchiris @e0ne @ykulazhenkov @rollandf WDYT?

rollandf:
Agree with the approach

ykulazhenkov (approved):

PR #78, opened 2025-02-19
add NicFirmwareTemplate and NicFirmwareSource CRDs

almaslennikov:
Created a separate PR with CRD changes to facilitate merging


api/v1alpha1/nicfirmwaretemplate_types.go
```diff
@@ -0,0 +1,72 @@
...
+	// UpdatePolicy indicates whether the operator needs to validate installed FW or upgrade it
+	// +kubebuilder:validation:Enum=Validate;Update
+	// +required
+	UpdatePolicy string `json:"updatePolicy"`
+}
+
+// NicFirmwareTemplateSpec defines the FW templates and node/nic selectors for it
+type NicFirmwareTemplateSpec struct {
+	// NodeSelector contains labels required on the node
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
rollandf, line 34: Maybe use:
`NodeSelector *corev1.NodeSelector json:"nodeSelector,omitempty" `

rollandf (approved):
LGTM.
One suggestion about NodeSelector

api/v1alpha1/nicfirmwaretemplate_types.go
rollandf: I see here that we use the same map[string]string in the other template.

https://github.com/Mellanox/nic-configuration-operator/blob/main/api/v1alpha1/nicconfigurationtemplate_types.go#L93

Let's keep the same

api/v1alpha1/nicfirmwaresource_types.go
```diff
@@ -0,0 +1,60 @@
...
+*/
+
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// NicFirmwareSourceSpec represents a list of url sources for FW
+type NicFirmwareSourceSpec struct {
+	// BinUrlSources represents a list of url sources for FW
+	BinUrlSources []string `json:"binUrlSources"`
```
ykulazhenkov, line 23: Not sure what are our preferences about "strict" vs "relaxed" validation for this project, but we can validate that binUrlSources field has at lease one element

api/v1alpha1/nicfirmwaresource_types.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// NicFirmwareSourceStatus represents the status of the FW from given sources, e.g. version available for PSIDs
+type NicFirmwareSourceStatus struct {
+	// State represents the firmware processing state, e.g. Downloading|Processing|Success|ProcessingFailed|DownloadFailed
+	State string `json:"state"`
```
ykulazhenkov, line 29: do we want to add enum validation for this field?

api/v1alpha1/nicfirmwaresource_types.go
```diff
@@ -0,0 +1,60 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// NicFirmwareSourceSpec represents a list of url sources for FW
+type NicFirmwareSourceSpec struct {
```
ykulazhenkov, line 21: this applies to all types in the file.
In the FirmwareTemplate you explicitly using `// +required` tag. Should we add it for the required fields in this file as well?

api/v1alpha1/nicfirmwaretemplate_types.go
ykulazhenkov: let's document behavior for an empty node selector as a part of the docstring.

almaslennikov: Done

api/v1alpha1/nicfirmwaresource_types.go
almaslennikov: Added validation for non-empty array

api/v1alpha1/nicfirmwaresource_types.go
almaslennikov: Added enum validation

api/v1alpha1/nicfirmwaresource_types.go
almaslennikov: Added required tags to the required fields

ykulazhenkov (approved):

PR #79, opened 2025-02-19
Add a controller for NicFirmwareTemplates

almaslennikov:
The template apply logic is very similar to the one we already have for the NicConfigurationTemplate, so let's extract it into a separate file and reuse in both controllers


ykulazhenkov (approved):

e0ne (approved):
LGTM

PR #88, opened 2025-03-04
include stderr logs for easier debugging

kitfoman:
Root-cause is difficult to debug whenever any of the underlying utilities fail. This adds a change to include stderr of the executed command whenever it fails for most things in `utils.go`


rollandf:
Hey, thanks for the PR.
Can you please add signed-off in commit:
`Signed-off-by: Author Name <authoremail@example.com>`

PR #107, opened 2025-04-30
chore: Refactor internal host package

almaslennikov:
For the upcoming DMS integration, we need to restructure some internal packages for better code structure.

The host package is split into three:

devicediscovery - DiscoverNicDevices() function
host - only host related utils: ScheduleReboot(), GetOfedVersion, GetHostUptime()
configuration - most of the former host package will stay here with all the nic configuration related functionality

No new logic is added in this PR, it only consists of functions moved to new packages with their respective unit tests


ykulazhenkov (approved):
LGTM

PR #127, opened 2025-06-27
feat: add BFB update

almaslennikov:


internal/controller/nicfirmwaresource_controller.go
```diff
@@ -95,68 +95,133 @@ func (r *NicFirmwareSourceReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		}
+		return nil, err
+	}
+
+	log.Log.Info("BFB file successfully processed", "cacheName", instance.Name, "filename", bfbFileName)
+	return bfbVersions, nil
 }
 
-func (r *NicFirmwareSourceReconciler) updateStatus(ctx context.Context, obj *v1alpha1.NicFirmwareSource, status string, statusError error, versions map[string][]string) error {
+func (r *NicFirmwareSourceReconciler) updateStatus(ctx context.Context, obj *v1alpha1.NicFirmwareSource, status string, statusError error, versions map[string][]string, bfbVersions map[string]string) error {
```
heyvister1, line 224: I suggest changing versions to binVersions
```suggestion
func (r *NicFirmwareSourceReconciler) updateStatus(ctx context.Context, obj *v1alpha1.NicFirmwareSource, status string, statusError error, binVersions map[string][]string, bfbVersions map[string]string) error {
```

I know this will require API change to replace `Versions` to `BinaryVersions` WDYT? Since we're using v1alpha1 maybe its a good time to do that

Also updateStatus signature seems to be too long, mostly you call it with nil assignments. I suggest using `FirmwareVersions` struct to hold both `binVersions`, `bfbVersions` maps

pkg/firmware/manager.go
```diff
@@ -114,6 +128,46 @@ func (f firmwareManager) BurnNicFirmware(ctx context.Context, device *v1alpha1.N
 	fwSourceName := device.Spec.Firmware.NicFirmwareSourceRef
 
 	cacheDir := path.Join(f.cacheRootDir, fwSourceName)
+
```
heyvister1, line 131: `BurnNicFirmware` seems to be too long. I suggest breaking `BurnNicFirmware` into small functions. Handling BFB or Bin sources. Searching for binaries, preparing and coping binaries if needed.

rollandf (approved):
LGTM

internal/controller/nicfirmwaresource_controller.go
almaslennikov: Doesn't save much space actually, let's leave it like this

PR #130, opened 2025-07-20
fix: README Configuration details update

rollandf:


rollandf:
For better parsing from docs repo

PR #166, opened 2025-09-26
feat: provision, install and run DOCA SPC-X CC package

almaslennikov:
The package should be provisioned via NicFirmwareSource. Link to .deb package for Ubuntu 22.04 should be provided.

The CC algorithm is to be run on runtime configuration apply when SPC-X optimizations are enabled


api/v1alpha1/nicfirmwaresource_types.go
```diff
@@ -27,6 +27,9 @@ type NicFirmwareSourceSpec struct {
 	// BFBUrlSource represents a url source for BlueField Bundle
 	// +optional
 	BFBUrlSource string `json:"bfbUrlSource,omitempty"`
+	// DocaSpcXCCUrlSource represents a url source for DOCA SPC-X CC .deb package for ubuntu 22.04
```
rollandf, line 30: I assume this API will be deprecated along the way.
Should we add some comment about it?

internal/controller/nicdevice_controller.go
```diff
@@ -364,7 +366,16 @@ func (r *NicDeviceReconciler) applyRuntimeConfig(ctx context.Context, status *ni
 		}
 	}
 
-	err := r.ConfigurationManager.ApplyDeviceRuntimeSpec(status.device)
+	targetVersion, err := r.SpectrumXManager.GetDocaCCTargetVersion(status.device)
+	if err != nil {
+		return err
```
rollandf, line 371: Should we update the Condition here?

rollandf (approved):

PR #221, opened 2026-01-12
feat: add fedramp support

rollandf:


Dockerfile.nic-configuration-daemon.stig
```diff
@@ -0,0 +1,97 @@
...
+
+# Copy bindata directory to the root directory where the binary runs
+COPY bindata /bindata
+
+# Copy sources to the container
+COPY . /workspace
+
+RUN --mount=type=secret,id=stig_script,target=/tmp/stig-fixer.sh bash /tmp/stig-fixer.sh
+
+USER 65532:65532
```
almaslennikov: We should set the user to root for this container

almaslennikov (approved):

PR #246, opened 2026-01-27
feat: add support for RHEL stig images

almaslennikov:
also install only needed packages from DOCA
install MFT in operator images
Move to mlxvpd instead of mstvpd where it's not available


rollandf (approved):

maze88 (commented):
additional [nit] for uniformity, brevity, and clarity:
i suggest setting the Dockerfile suffixes to reflect the GitLab CI build stage names "operator" & "daemon"

heyvister1 (approved):

maze88 (approved):

PR #319, opened 2026-03-20
feat: add RoCE mode, ECN, pause frames, cable length, and runtime perf tuning

synthe102:
Extend NIC configuration with new runtime settings:
- RoCE mode (v1/v2) via cma_roce_mode
- ECN (Explicit Congestion Notification) via sysfs
- Global pause frame control via ethtool
- Cable length for ECN buffer thresholds via mlnx_qos
- Runtime performance tuning: ring buffers, combined channels, LRO via ethtool

Refactor CalculateDesiredRuntimeConfig to return a DesiredRuntimeConfig struct instead of multiple return values, and use consts.TrustModeDscp for defaults.


almaslennikov:
Hi @synthe102! We are now in the process of stabilizing the next release of the Network Operator, and for this feature we still need to figure out the QA scope. So we'll revisit this after the 26.4.0 version is released. Sorry for the delay, will keep you posted

pkg/configuration/utils.go
```diff
@@ -241,6 +280,355 @@ func (h *configurationUtils) SetQoSSettings(device *v1alpha1.NicDevice, spec *v1
...
+			txFound = true
+			txOn = strings.Contains(lower, "on")
+		}
+	}
+	if !rxFound || !txFound {
+		return false, fmt.Errorf("RX/TX pause frame status not found in ethtool -a output for %s", interfaceName)
+	}
+	// Use || so that a partial state (e.g. rx=off, tx=on) is treated as "enabled",
+	// ensuring Enabled=false triggers a full disable even from a partial state.
+	return rxOn || txOn, nil
```
almaslennikov: This is correct for Enabled: false (partial-on ⇒ re-apply), but wrong for Enabled: true: if RX gets externally disabled while TX stays on, actual == true and the validator concludes the desired state is applied — drift goes undetected, which is the operator's core guarantee.

```suggestion
  if qos.PauseFrames.Enabled {
      if !(rxOn && txOn) { return false, nil }
  } else {
      if rxOn || txOn { return false, nil }
  }
```

pkg/configuration/utils.go
```diff
@@ -241,6 +280,355 @@ func (h *configurationUtils) SetQoSSettings(device *v1alpha1.NicDevice, spec *v1
...
+// nsenterWithFlags wraps a command with nsenter using the specified namespace flags.
+func (h *configurationUtils) nsenterWithFlags(nsFlags []string, args ...string) execUtils.Cmd {
+	fullArgs := append([]string{"-t", "1"}, nsFlags...)
+	fullArgs = append(fullArgs, "--")
+	fullArgs = append(fullArgs, args...)
+	return h.execInterface.Command("nsenter", fullArgs...)
+}
+
+// nsenterCommand wraps a command with nsenter to run in the host's network namespace.
+func (h *configurationUtils) nsenterCommand(args ...string) execUtils.Cmd {
```
almaslennikov: Why do we need to wrap the commands in `nsenter`?

pkg/configuration/utils.go
```diff
@@ -241,6 +280,355 @@ func (h *configurationUtils) SetQoSSettings(device *v1alpha1.NicDevice, spec *v1
...
+
+// GetCombinedChannels returns the current number of combined channels
+func (h *configurationUtils) GetCombinedChannels(interfaceName string) (int, error) {
+	log.Log.V(2).Info("ConfigurationUtils.GetCombinedChannels()", "interface", interfaceName)
+	cmd := h.nsenterCommand("ethtool", "-l", interfaceName)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		// ethtool -l may fail (with or without error output) on drivers that don't support channel config
+		log.Log.V(2).Info("GetCombinedChannels(): ethtool -l failed, driver may not support it", "interface", interfaceName, "err", err)
+		return 0, nil
```
almaslennikov, line 532: ethtool failures, unparseable output, and "driver doesn't expose channels" all return (0, nil). The validator and apply path both treat actual == 0 as "skip this knob". A transient ethtool error therefore silently passes validation and skips re-apply on a driver that actually does support combined channels.

Let's at least log the output

pkg/configuration/utils.go
synthe102: Actually not needed since we run in `hostPID: true` and `hostNetwork: true` if deployed through helm. Will remove.

pkg/configuration/utils.go
synthe102: Thanks ! I went a slightly different route and return `rx, tx` from `GetPauseFrames()` instead.

almaslennikov (approved):
@synthe102 LGTM, thank you for the contribution! Sorry it dragged for a while

PR #320, opened 2026-03-25
Remove control plane tolerations

kreeuwijk:
Remove control plane tolerations

Remove unneeded tolerations that can cause the pod to run on a node without a CSI.
The NIC Config Operator only typically runs either on worker nodes, or on nodes that act as both workers and CP nodes. In both of these cases, the node doesn't have a control plane taint. Hence this toleration is not desirabe.

We have seen a customer use case where the CSI driver isn't installed on the CP nodes, which then causes problems for the operator pod if it gets scheduled on a CP node, since it cannot access the firmware PVC. Removing the tolerations fixes that.

### DIFF
--- a/deployment/nic-configuration-operator-chart/values.yaml
+++ b/deployment/nic-configuration-operator-chart/values.yaml
@@ -6,13 +6,7 @@ operator:
     # -- image tag to use for the operator image
     tag: latest
   # -- tolerations for the operator
-  tolerations:
-    - key: "node-role.kubernetes.io/master"
-      operator: "Exists"
-      effect: "NoSchedule"
-    - key: "node-role.kubernetes.io/control-plane"
-      operator: "Exists"
-      effect: "NoSchedule"
+  tolerations: []
   # -- node selector for the operator
   nodeSelector: {}
   # -- node affinity for the operator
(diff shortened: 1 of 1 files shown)


almaslennikov:
I think the fix should be in the network operator repo, since that's the deployment model we support. The helm chart in this repo is only for reference.

What we could do is:
1) extend the NicClusterPolicy's section for nic operator with deployment tolerations and nodeaffinity (like how it's done now on the CR level: https://github.com/Mellanox/network-operator/blob/master/api/v1alpha1/nicclusterpolicy_types.go#L436).
2)  in the manifest template delete the hardcoded parts: https://github.com/Mellanox/network-operator/blob/master/manifests/state-nic-configuration-operator/060-operator.yaml#L21
3) Add those default tolerations and nodeaffinity to the example of NIC Cluster Policy we use in the docs: https://github.com/Mellanox/network-operator/blob/master/hack/templates/crs/mellanox.com_v1alpha1_nicclusterpolicy_cr-full.template#L114

That way, we are not changing the default behavior too much for all users and allowing those who need to customize their deployments

kreeuwijk:
That's fair, thank you

almaslennikov:
@kreeuwijk can we close this one?

kreeuwijk:
Yes, please ping me here when the NicClusterPolicy has been updated to support it instead.

kreeuwijk:
@almaslennikov any update on when this ships?

PR #351, opened 2026-04-24
Identify NICs by PCI device address instead of serial number

almaslennikov:
On HGX B300, multiple ConnectX NICs are soldered onto a single baseboard and flashed with the same VPD image during NVIDIA manufacturing, so they report identical serial numbers. The operator was keying ports-into- devices, the DMS client map, and CR names by SN, collapsing all embedded cards into a single NicDevice and routing every card's DMS calls to one client (last-write-wins).

Switch the uniqueness key to the PCI device address (Domain:Bus:Device, function stripped) across discovery, the reconciler, DMS client lookup, and CR naming. Ports of the same physical card share Dom:Bus:Dev and differ only by function, so the key naturally groups multi-function NICs (including 4-PF ConnectX-9 cards) while staying unique across cards on B300.

- pkg/utils: new PCIDeviceAddress helper
- pkg/devicediscovery: return map keyed by PCI device
- controller: sanitize PCI key for DNS-1123 CR names (co-node-25-101b-0000-04-00)
- pkg/dms: rename GetDMSClientBySerialNumber → GetDMSClientByPCIAddress; add GetDMSClientForDevice helper
- api: keep Status.SerialNumber populated (informational); note non-uniqueness on SerialNumber/SerialNumbers field comments

### DIFF
--- a/internal/controller/devicediscovery_controller.go
+++ b/internal/controller/devicediscovery_controller.go
@@ -37,6 +37,7 @@ import (
 	"github.com/Mellanox/nic-configuration-operator/pkg/devicediscovery"
 	"github.com/Mellanox/nic-configuration-operator/pkg/helper"
 	"github.com/Mellanox/nic-configuration-operator/pkg/host"
+	"github.com/Mellanox/nic-configuration-operator/pkg/utils"
 )
 
 var deviceDiscoveryReconcileTime = time.Minute * 5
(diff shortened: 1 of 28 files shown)


internal/controller/devicediscovery_controller.go
```diff
@@ -52,9 +53,15 @@ type DeviceDiscoveryController struct {
 	namespace       string
 }
 
-// Constructs a unique CR name based on the device's type and serial number
-func (d *DeviceDiscoveryController) getCRName(deviceType string, serialNumber string) string {
-	return strings.ToLower(d.nodeName + "-" + deviceType + "-" + serialNumber)
+// crNameSanitizer replaces PCI-address punctuation (:, .) with DNS-1123-safe dashes.
+var crNameSanitizer = strings.NewReplacer(":", "-", ".", "-")
```
rollandf, line 57: What will happen on upgrade when we have exisiting CRs with previous naming?

e0ne (approved):

rollandf (approved):
LGTM.
Only question is what is the behavior on upgrade when we have existing CRs.

internal/controller/devicediscovery_controller.go
almaslennikov: Nothing, the name of the CR is only for display to the user. The operator will match the device based on its PCI address. If according to the new logic, the amount of ports for a device will shrink, new CRs will be created.

At worst, it will lead to inconsistent naming on a node, which shouldn't be a big deal

PR #366, opened 2026-05-20
ci: disable pulse scan-image component usage tracking

rollandf:
The scan-image component auto-generates a *-track-usage telemetry job that requires a 'generic' runner tag not available in this project, causing it to be stuck indefinitely. Set track_usage=false on all 4 component includes to suppress these jobs. No impact on scan behavior.


e0ne (approved):
LGTM

PR #375, opened 2026-06-01
Add ConnectX-9 Network Bay (set_system_conf) support and force knob

almaslennikov:
ConnectX-9 Network Bay ("orchid") cards expose two CX9 ASICs as two PCI endpoints that share a single OSFP cage and must be configured as a pair using a new per-ASIC knob, mlxconfig set_system_conf <conf>[<asic>].

This adds:
- A networkBay template block (CEL-gated to nicType 1025) and a global force knob (--force on mlxconfig) on ConfigurationTemplateSpec.
- Orchid auto-detection in device discovery: GetNetworkBayASIC reads the MGIR ga/ga_valid register fields via mlxreg and surfaces the ASIC index on NicDeviceStatus.networkBay; siblings are paired by serial number to populate PeerPCI. Detection never fails discovery.
- nvconfig wrappers SetSystemConf / ValidateSystemConf, and a force flag on SetNvConfigParametersBatch.
- A whole-bay template-matching constraint: a networkBay template is only applied to a node when its matched devices form complete bays (even count, every device detected as part of a bay, serials pairing into twos); otherwise the node is rejected with device + template conditions and a warning event.
- set_system_conf / validate_system_conf layered onto the apply/validate flow after the regular config and after the Spectrum-X path converges. The Spectrum-X breakout/postBreakout flow is intentionally left untouched (reworked in a separate FR).

Covered by unit tests and a revised mermaid reconcile diagram replacing the outdated PNG.


rollandf (approved):
LGTM

PR #384, opened 2026-06-30
configuration: stop managing ADVANCED_PCI_SETTINGS

almaslennikov:
## Summary

The configuration manager currently auto-enables `ADVANCED_PCI_SETTINGS` when it's off, runs `mlxfwreset` to avoid a reboot, and re-queries nvconfig to pick up newly visible params. The validation path also rejects any spec referencing a param hidden because ADVANCED is off, returning `IncorrectSpecError`.

That entangles the operator with a platform-policy knob better owned by the deployer (or by the device template via `rawNvConfig`). This PR drops the auto-enable entirely. Hidden params are now treated as unsupported and skipped — matching the existing `hasUnsupportedParams` branch in apply (`ApplyStatusPartiallyApplied`).

`consts.AdvancedPCISettingsParam` is kept so users can still drive the param explicitly via `template.rawNvConfig`.

## Changes

* `ApplyNVConfiguration`: drop the auto-enable + `mlxfwreset` + re-query block.
* `ValidateDeviceNvSpec`: drop the `advancedPciSettingsEnabled` guard and the `IncorrectSpecError` branch; treat missing-from-`NextBootConfig` as a skip (no `configUpdateNeeded`).
* `applyResetToDefault`: drop the post-reset `ADVANCED_PCI_SETTINGS=true` re-set.
* `configValidation.ConstructNvParamMapFromTemplate`: drop the ADVANCED gate around the MaxAccOutRead default-derivation; skip silently when the param is hidden instead of returning `IncorrectSpecError`.
* `configValidation.ValidateResetToDefault`: collapse to BF3-mode preservation plus plain `reflect.DeepEqual` between Current/NextBoot and Default.
* Remove `AdvancedPCISettingsEnabled` from the `configValidation` interface, impl, and mock; rework Ginkgo specs accordingly.

## Test plan

- [x] `go build ./...`
- [x] `go test ./pkg/configuration/... -count=1` — pass
- [x] `go test ./... -count=1` — all unit/integration suites pass (the one flake in `internal/controller` cleared on rerun; `test/e2e` failure is a sandboxed `kubectl create -f https://github.com/...`, unrelated)
- [x] `golangci-lint run ./pkg/configuration/...` — 0 issues
- [ ] E2E on a NIC where `ADVANCED_PCI_SETTINGS` is off and the template references a hidden param: reconcile should complete and report `ApplyStatusPartiallyApplied`, without auto-enabling ADVANCED.

### DIFF
--- a/api/v1alpha1/nicconfigurationtemplate_types.go
+++ b/api/v1alpha1/nicconfigurationtemplate_types.go
@@ -126,7 +126,7 @@ type GpuDirectOptimizedSpec struct {
 type SpectrumXOptimizedSpec struct {
 	// Optimize Spectrum X
 	Enabled bool `json:"enabled"`
-	// Version of the Spectrum-X architecture to optimize for. Should match the name of the config map with Spectrum-X profile 
+	// Version of the Spectrum-X architecture to optimize for. Should match the name of the config map with Spectrum-X profile
 	// +required
 	Version string `json:"version"`
 	// Overlay mode to be configured
--- a/internal/controller/nicdevice_controller.go
+++ b/internal/controller/nicdevice_controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"strings"
 	"sync"
 	"time"
 
@@ -87,6 +88,9 @@ type nicDeviceConfigurationStatus struct {
 	requestedFirmwareVersion     string
 	nvConfigUpdateRequired       bool
 	rebootRequired               bool
+	// unsupportedNvParams lists desired nv params that are hidden on this device (absent from
+	// NextBootConfig). Surfaces as PartiallyApplied in the ConfigUpdateInProgress condition.
+	unsupportedNvParams []string
 }
 
 // Reconcile reconciles the NicConfigurationTemplate object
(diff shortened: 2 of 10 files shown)


almaslennikov:
Addressed both Greptile findings (force-pushed `cfaaaa3`):

1. **Stale `ENABLE_PCI_OPTIMIZATIONS` comment** — `pkg/configuration/configvalidation.go:68` now references `ADVANCED_PCI_SETTINGS`, the current knob.

2. **All-hidden-params observability gap** — `ConfigurationManager.ValidateDeviceNvSpec` now returns a 4-tuple `(bool, bool, []string, error)`; the new slice carries sorted, deduped names of desired params that are hidden on the device. The controller threads them through `nicDeviceConfigurationStatus.unsupportedNvParams` and, in `applyRuntimeConfig`, promotes the final `ConfigUpdateInProgress` reason from `UpdateSuccessful` to a new `PartiallyApplied` (defined in `pkg/consts/consts.go`) with a message listing the skipped param names. This surfaces correctly even when `ApplyNVConfiguration` is never invoked because all desired params were hidden. Mock + tests updated to match the new signature; new assertions in `pkg/configuration/manager_test.go` cover the slice return.

`go test ./pkg/configuration/... ./internal/controller/...` and `golangci-lint run` pass cleanly.

PR #385, opened 2026-06-30
configuration: apply Network Bay system_conf before overrides

almaslennikov:
## Summary

`mlxconfig set_system_conf <conf>[<asic>]` applies a **named profile** — a fixed batch of nv params (`BOARD_CONFIGURATION_MODE`, `MODULE_SPLIT_M0[*]`, `LINK_TYPE_P*`, `NUM_OF_PLANES_P*`, `NUM_OF_PF`, …). It was previously layered **last**, after the regular nv-config batch and after the Spectrum-X breakout/postBreakout path. For Network Bay devices that also carry `rawNvConfig` overrides, that ordering broke two ways:

- `set_system_conf` ran **after** the overrides were staged, silently overwriting any `rawNvConfig` param that belongs to the profile — the user's override was lost.
- `validate_system_conf` was collapsed to a single match/no-match bit, so a `rawNvConfig` override of a profile param read as permanent drift and **looped reboots** on every reconcile.

This PR makes `set_system_conf` the **baseline, applied first**, with the rest layering on top in the same next-boot config.

## Changes

- **Reorder** (`pkg/configuration/manager.go`): `ApplyNVConfiguration` runs `applySystemConf` at the very top (before the Spectrum-X and regular branches); `ValidateDeviceNvSpec` and `validateSpectrumXNvSpec` check `systemConfMismatch` first and short-circuit. Every later stage (Spectrum-X / regular / `rawNvConfig`) overrides on top → overrides win, single-reboot convergence.
- **Per-param parse** (`pkg/nvconfig/utils.go`): add `ValidateSystemConfDetailed` returning the OK/MISMATCH/SKIPPED rows of `validate_system_conf` (the data was always emitted; only the trailing `Result:` line was consumed). `ValidateSystemConf` stays as a thin wrapper for library consumers.
- **Override-aware drift** (`systemConfMismatch`): MISMATCH rows whose param name is in `template.rawNvConfig` are treated as intentional overrides, not drift. Matching is by **name only** (robust to `validate_system_conf` value encodings like `LINK_TYPE_P1=ETH` vs the numeric raw value); the override's value is still verified by the regular nv param validation. `SKIPPED` rows are informational and never drive apply.

No CRD / API / Helm changes.

## Test plan

- [x] `go build ./...`
- [x] `go test ./pkg/nvconfig/... -count=1` — new `parseValidateSystemConfDetailed` specs (all-OK conf6, mixed MISMATCH/OK conf3, no-`Result`-line fallback, garbage)
- [x] `go test ./pkg/configuration/... -count=1` — Network Bay apply/validate reworked for the new ordering, plus override-explained (no apply), unexplained (apply), and `set_system_conf`-before-batch ordering specs
- [x] `go test ./internal/controller/... -count=1`
- [x] `golangci-lint run ./...` → 0 issues
- [ ] E2E on a ConnectX-9 Network Bay node: first reconcile applies `set_system_conf`, reboots; adding a `rawNvConfig` override of a profile param (e.g. `NUM_OF_PF`) applies the override and converges with no reboot loop; out-of-band drift of a non-overridden profile param re-applies `set_system_conf`.

Follow-up to #384.


almaslennikov:
Addressed the review (force-pushed `7cf11bf`):

1. **ResetToDefault + NetworkBay reboot loop (the 3/5 blocker)** — real bug, fixed. `hasNetworkBaySpec` now excludes devices with `ResetToDefault: true`, so `applySystemConf`/`systemConfNeedsApply` short-circuit and the reset path runs alone. Without it, apply would stage `set_system_conf` and the reset would wipe it every reconcile. Added controller-free specs covering both validate and apply for `ResetToDefault + NetworkBay` (asserts `ValidateSystemConfDetailed`/`SetSystemConf` are never called).

Also extended the fix beyond what the review covered, per maintainer feedback: the Spectrum-X profile can override system_conf too, not just `rawNvConfig`. Override priority is now **rawNvConfig > Spectrum-X > system_conf**:

- `systemConfMismatch` was split into `systemConfMismatchedParams` (returns the MISMATCH param names) + `systemConfNeedsApply`, which treats a mismatch as real drift only when the param is not in `higherPriorityNvParams` — `rawNvConfig` params **plus**, when Spectrum-X is enabled, the breakout + postBreakout params.
- Apply order is `set_system_conf` → Spectrum-X → `rawNvConfig` (raw merged last so it wins). Added a NetworkBay+Spectrum-X spec proving a breakout-owned mismatch isn't treated as system_conf drift.

On the two P2 parser notes (`parseValidateSystemConfDetailed`): both only arise on malformed input — real `validate_system_conf` output terminates each section with a `---` separator or the `Result:` line, which resets the skipped-section state, and OK rows always contain `=`. I kept the parser lenient (skip unrecognized lines) rather than erroring on partial input; happy to tighten if you'd prefer it reject malformed rows.

In-repo docs updated: regenerated `docs/api-reference.md` (now documents `NetworkBaySpec`/`NicDeviceNetworkBayStatus`) and refreshed `docs/nic-configuration-library.md` (NV-config flow + `ValidateSystemConfDetailed`).

`go build`, `go test ./pkg/configuration/... ./pkg/nvconfig/...`, and `golangci-lint run ./...` all clean. (The one `internal/controller` flake is pre-existing and unrelated — Phase 2 doesn't touch that package; it passes on rerun.)

almaslennikov:
Reworked the validate/apply structure (force-pushed `5854340`) so the system_conf logic is one explicit flow, not split across helpers. Supersedes my earlier comment describing `systemConfNeedsApply`/`higherPriorityNvParams` — those are gone.

**`ValidateDeviceNvSpec` is now a single linear flow:**
1. query nv config (all ports)
2. reset-to-default → return
3. `systemConfMismatchedParams` → the params that don't match the named profile
4. build the higher-priority override config: Spectrum-X breakout+postBreakout (rawNvConfig merged) *or* template+rawNvConfig
5. **system_conf drift, inline:** each mismatched param not covered by that override config → update+reboot (priority `rawNvConfig` > Spectrum-X > system_conf)
6. validate the override params' values against the device

`validateSpectrumXNvSpec` was folded in and deleted. The only extracted helpers are the pure param-**value** checks (`spectrumXParamsNeedUpdate`, `validateTemplateParamsApplied`) — the system_conf decision stays inline (kept under gocyclo by moving just the value-validation out).

**`ApplyNVConfiguration` just applies** — it does not re-derive drift. `ValidateDeviceNvSpec` already decided an update is needed; apply queries → handles reset → (Spectrum-X or regular), and `applySystemConf` unconditionally (re)stages `set_system_conf` as the baseline before the override params. No `validate_system_conf` call on the apply side.

`ResetToDefault` still excluded from Network Bay management (`hasNetworkBaySpec`) — addresses the earlier reboot-loop finding.

`go build`, `go test ./pkg/configuration/... ./pkg/nvconfig/... ./internal/controller/...`, and `golangci-lint run ./...` all clean.

almaslennikov:
Reworked validate + apply into the single explicit flow (force-pushed `5c51d4e`), and added logging.

**`ValidateDeviceNvSpec` — one flow (each step's mismatch fails validation):**
1. query nv config
2. reset-to-default → return
3. `validate_system_conf` → note mismatched profile params
4. Spectrum-X PRE breakout (+ rawNvConfig overlaid) → value-check
5. Spectrum-X POST breakout (+ rawNvConfig overlaid) → value-check
6. desired config from spec template (+ all rawNvConfig) → value-check
7. coverage: if PRE breakout + POST breakout + desired + rawNvConfig do **not** cover every system_conf mismatch, the baseline drifted → set_system_conf re-apply required

**`ApplyNVConfiguration` — just applies:**
1. query → 2. reset (+apply, finish) → 3. note system_conf mismatches → 4. build the combined `breakout + postBreakout + desired + rawNvConfig` map (raw wins) → 5. if the map doesn't cover all mismatches, apply `set_system_conf` → 6. `force==true`: apply **all** combined params (`mlxconfig --force`) → 7. `force==false`: apply only the combined params that exist in the query and differ (which naturally sequences breakout before postBreakout across reboots).

Priority throughout: **rawNvConfig > Spectrum-X / template > system_conf**.

Removed the old two-branch structure, `applySpectrumXNVConfiguration`, and the per-branch system_conf checks. `validateSpectrumXNvSpec` is gone (folded in). Added meaningful default-level logs (system_conf drift, set_system_conf apply, force apply, final result) and `V(2)` debug logs (mismatched params, the breakout/postBreakout/desired maps, the combined map).

`go build`, `go test ./pkg/configuration/... ./pkg/nvconfig/... ./internal/controller/...`, and `golangci-lint run ./...` all clean.

pkg/configuration/manager.go
```diff
@@ -104,13 +100,112 @@ func (h configurationManager) ValidateDeviceNvSpec(ctx context.Context, device *
...
+		mergeParams(overrides, postBreakoutParams)
+		log.Log.V(2).Info("validating Spectrum-X params", "device", device.Name, "breakout", breakoutParams, "postBreakout", postBreakoutParams)
+
+		needUpdate, err := h.spectrumXParamsNeedUpdate(ctx, device.Status.Ports[0].PCI, breakoutParams, postBreakoutParams)
+		if err != nil {
+			return false, false, nil, err
+		}
+		if needUpdate {
+			log.Log.Info("Spectrum-X breakout/postBreakout config not yet applied", "device", device.Name)
+			configUpdateNeeded = true
```
almaslennikov: [P1] Keep staged Spectrum-X changes rebootable

`spectrumXParamsNeedUpdate` compares against `CurrentConfig` and maps any mismatch to `configUpdateNeeded=true`, while `ApplyNVConfiguration` later diffs against `NextBootConfig`. With current `NUM_OF_PF=1`, next boot `=2`, and desired `=2`, validation requests an update and reboot, apply skips the already-staged value and returns `NothingToDo` with `RebootRequired=false`, and `applyNvConfig` overwrites the validation reboot flag with that false result. Reconciliation then repeats without ever booting the staged value. Please split Spectrum-X validation into next-boot update versus current-state reboot, as the regular path does, or otherwise preserve the reboot result, and cover this transition.

pkg/configuration/manager.go
```diff
@@ -192,72 +264,139 @@ func (h configurationManager) ApplyNVConfiguration(ctx context.Context, device *
...
+		//    them — this is what sequences breakout before postBreakout without --force.
+		for pciAddr, nvConfig := range nvConfigsForPorts {
+			supportedParams := map[string]string{}
+			for param, value := range desiredParams {
+				nextValues, found := nvConfig.NextBootConfig[param]
+				if !found {
+					hasUnsupportedParams = true
+					continue
+				}
+				if !slices.Contains(nextValues, value) {
```
almaslennikov, line 269: [P1] Reapply overrides after rewriting the baseline

`nvConfigsForPorts` was captured before `SetSystemConf`. Suppose raw `NUM_OF_PF=8` already matches while an unrelated `BOARD_CONFIGURATION_MODE` mismatch triggers `set_system_conf`. That command restages the whole profile and overwrites `NUM_OF_PF`, but this loop consults the stale pre-call snapshot, decides `NUM_OF_PF` still matches, and skips the override. The first reboot therefore loses the user value and a second apply/reboot is needed. When `systemConfApplied` is true, re-query or reapply all desired overrides after the baseline.

pkg/configuration/manager.go
```diff
@@ -192,72 +264,139 @@ func (h configurationManager) ApplyNVConfiguration(ctx context.Context, device *
...
-			nextValues, found := nvConfig.NextBootConfig[param]
-			if !found {
-				log.Log.Info("Parameter not found in NextBootConfig, skipping", "device", device.Name, "param", param)
-				hasUnsupportedParams = true
+	if options.Force {
+		// 6. force: apply every combined param in one batch (mlxconfig --force accepts params that are
+		//    not currently visible, e.g. per-port params before a breakout reboot exposes their ports).
+		if len(desiredParams) > 0 {
+			log.Log.Info("applying all nv config params with --force", "device", device.Name, "count", len(desiredParams))
+			if err := h.nvConfigUtils.SetNvConfigParametersBatch(firstPortPCIAddr, desiredParams, options.WithDefault, true); err != nil {
```
almaslennikov: [P1] Apply forced batches to every PCI target

This branch sends the batch only to `firstPortPCIAddr`, whereas the non-force branch and the previous implementation apply per PCI. On a multi-PF `NicDevice` with `Force=true`, if PF2 needs `TRACER_ENABLED` but PF1 does not, each reconcile writes PF1, reports a reboot, and never changes PF2, producing a reboot loop. The forced path still needs to iterate `nvConfigsForPorts` and target every PF that the template manages.

pkg/configuration/manager.go
```diff
@@ -192,72 +264,139 @@ func (h configurationManager) ApplyNVConfiguration(ctx context.Context, device *
...
+				log.Log.Error(err, "Failed to apply nv config parameters", "device", device.Name, "params", desiredParams)
+				return &types.ConfigurationApplyResult{Status: types.ApplyStatusFailed}, err
+			}
+			anyParamsApplied = true
+		}
+	} else {
+		// 7. non-force: apply only the combined params that exist in the query (per port) and whose value
```
almaslennikov: [P1] Preserve the Spectrum-X breakout reboot boundary

Flattening breakout and postBreakout into one map changes the required phase state machine. The old path explicitly applied breakout, returned for a reboot, then applied postBreakout, and always passed `force=false`. Here any postBreakout key already visible before breakout is batched against the old topology; the force branch sends even currently hidden dependent keys. A profile with `NUM_OF_PF` in breakout and `LINK_TYPE_P1` or `SRIOV_EN` in postBreakout can therefore be rejected or staged against the pre-breakout PF layout. Please keep breakout -> reboot -> postBreakout and insert the system-conf baseline without collapsing those phases.

pkg/configuration/manager.go
```diff
@@ -701,21 +725,31 @@ func getRawNvConfigParams(device *v1alpha1.NicDevice) map[string]string {
...
+		if _, ok := target[k]; ok {
+			target[k] = v
+		}
+	}
+}
+
+// allParamsPresent reports whether every key is present in m. An empty keys slice is vacuously true.
+func allParamsPresent(m map[string]string, keys []string) bool {
+	for _, k := range keys {
+		if _, ok := m[k]; !ok {
```
almaslennikov: [P1] Normalize indexed ranges before the coverage check

This exact-key lookup does not recognize that an accepted raw key such as `MODULE_SPLIT_M0[0..3]` covers detailed mismatch rows named `MODULE_SPLIT_M0[0]` through `[3]`. MFT system profiles use range syntax and expand those ranges into individual `ParamView` names in `validate_system_conf`. The manager therefore reapplies `set_system_conf`, force-applies the range override, reboots, and repeats forever because the individual mismatch names are never considered covered. Please expand/canonicalize range keys for coverage, or reject ranges explicitly, and add a range-override test.

pkg/nvconfig/utils.go
```diff
@@ -53,8 +53,41 @@ type NVConfigUtils interface {
 	// ValidateSystemConf reports whether the device's applied configuration matches the named system
 	// configuration for the given ASIC via `mlxconfig -d <pci> -y validate_system_conf <conf>[<asic>]`.
 	ValidateSystemConf(ctx context.Context, pciAddr string, conf string, asic int) (bool, error)
+	// ValidateSystemConfDetailed runs the same `validate_system_conf` command but returns the per-param
+	// OK/MISMATCH results plus the skipped params, instead of collapsing to a single match bit. Callers
+	// that allow explicit overrides (e.g. rawNvConfig) on top of a named system conf use this to decide
+	// whether a reported mismatch is intentional (overridden) or a real drift requiring re-apply.
+	ValidateSystemConfDetailed(ctx context.Context, pciAddr string, conf string, asic int) (*SystemConfValidationResult, error)
```
almaslennikov, line 67: [P1] Keep the v1 `NVConfigUtils` interface source-compatible

This repository is tagged through v1.3.1 and documents `pkg/` as a reusable Go library. Adding a required method to the exported interface makes every downstream fake, wrapper, or partial adapter fail to compile, including consumers that only inject it for `QueryNvConfig`. Retaining the old method does not preserve implementor compatibility. Please expose detailed validation through a separate extension interface/type assertion with a safe fallback, or treat this as an explicit v2 module-path change.

pkg/nvconfig/utils.go
```diff
@@ -267,52 +300,129 @@ func (h *nvConfigUtils) SetSystemConf(ctx context.Context, pciAddr string, conf
...
 	log.Log.V(2).Info("command output", "command", "mlxconfig validate_system_conf", "pciAddr", pciAddr, "output", string(output))
 
 	// mlxconfig validate_system_conf exits non-zero (e.g. 3) when the device configuration does
 	// NOT match the system conf — that is a valid result, not a command failure. Treat the parsed
-	// "Result:" line as authoritative regardless of exit code, and only surface the command error
-	// when no result line could be parsed (i.e. mlxconfig genuinely failed to run).
-	matches, parseErr := parseValidateSystemConf(output)
+	// output as authoritative regardless of exit code, and only surface the command error when the
+	// output couldn't be parsed at all (i.e. mlxconfig genuinely failed to run).
+	result, parseErr := parseValidateSystemConfDetailed(output)
```
almaslennikov: [P2] Do not suppress failures after partial validation output

The producer prints the trailing `Result:` only after all parameter queries complete; an exception returns before that line. The new fallback accepts any parsed row without `Result:`, so if `mlxconfig` prints one `OK:` row and is then killed or errors, `parseErr` is nil and this code discards the nonzero command error, potentially reporting a match from incomplete data. At minimum propagate `err` when no result line was parsed; add a partial-output plus nonzero-exit test.

almaslennikov (commented):
I found several correctness issues in the new combined NV-configuration flow. Inline comments include concrete failing state transitions and the missing coverage.

pkg/configuration/manager.go
almaslennikov: Keeping the phases collapsed, deliberately. The two-phase boundary existed because postBreakout params couldn't be *set* until breakout had taken effect — but both paths now handle that without an explicit phase:

- **force=false**: the per-PF loop filters to params present in `NextBootConfig`, so a postBreakout param whose port doesn't exist yet is skipped this round and picked up after the breakout reboot exposes it. It self-sequences to breakout→reboot→postBreakout, same convergence as before.
- **force=true**: the whole interdependent batch goes through `mlxconfig --force`, which exists precisely to "accept a batch it would otherwise refuse due to implicit parameter dependencies" — i.e. staging breakout+postBreakout together for next boot in one reboot.

The specific failure you describe (postBreakout batched against the old topology) is what the visibility filter prevents in non-force; and your related #1 finding — validation comparing `CurrentConfig` while apply diffs `NextBootConfig` — is fixed by routing breakout/postBreakout through the same next-boot validation as the regular params, so they no longer disagree. The one residual assumption is that `mlxconfig --force` can stage a post-breakout param for next boot before the breakout reboot; if that's false on this firmware, the force path (not the structure) is what would need a guard. Happy to add the explicit phase back if you've seen `--force` reject that on hardware.

pkg/configuration/manager.go
almaslennikov: Fixed. Validation no longer uses `checkMlxConfigMismatch`/`CurrentConfig` for breakout/postBreakout — they go through the same next-boot per-port path as the regular params (`validateTemplateParamsApplied` over the combined override map). So validation and apply now read the same source: a value already staged for next boot but not yet current reports `rebootNeeded=true, configUpdateNeeded=false`, the reboot flag survives, and the staged value actually boots instead of looping. Added a regression test ("requires reboot (not update) when a param is staged for next boot but not yet current").

almaslennikov:
Addressed the review (force-pushed `1f7dca3`). Summary of each finding:

**Fixed**
- **#1 (manager.go:145) — staged-not-rebooted reboot loop.** Breakout/postBreakout now validate against next-boot (same path as regular params), so validation and apply agree and the reboot flag survives. Regression test added. (replied inline)
- **#2 (manager.go:324) — stale snapshot after `set_system_conf`.** Apply re-queries `nvConfigsForPorts` after `set_system_conf` so overrides aren't skipped against the pre-baseline snapshot.
- **#3 (manager.go:306) — forced batch only hit the first PF.** The force path now iterates every PF in `nvConfigsForPorts`.
- **#5 (manager.go:748) — range-key coverage.** Coverage now range-expands override keys (`MODULE_SPLIT_M0[0..3]` → `[0]`..`[3]`) before checking against the per-index `validate_system_conf` mismatch rows. Range-override test added.
- **#6 (utils.go:60) — v1 interface compat.** `ValidateSystemConfDetailed` moved out of `NVConfigUtils` into a separate optional `SystemConfValidator` interface; the manager type-asserts with a safe fallback (skips Network Bay management if absent). Existing `NVConfigUtils` implementors keep compiling.
- **utils.go:390 (P2)** — `OK:` rows without `=` are now kept (empty Actual) instead of silently dropped.
- **utils.go:380 (P2)** — parser resets the SKIPPED state on any unrecognized line (`default:` arm), so a header after a SKIPPED block can't be mis-captured.
- **utils.go:324 (P2)** — when no `Result:` line is parsed *and* the command errored, the command error is now propagated instead of trusting a derived match from partial output.

**Defended (not changed)**
- **#4 (manager.go:313) — keep breakout/postBreakout collapsed.** Replied inline: force=false self-sequences via the next-boot visibility filter; force=true relies on `mlxconfig --force` (its documented purpose) to stage the interdependent batch; #1's fix removes the validate/apply disagreement. Only the force path would need a guard, and only if `--force` can't stage a post-breakout param pre-reboot on this firmware — flagged that open question inline.

`go build`, `go test ./pkg/configuration/... ./pkg/nvconfig/... ./internal/controller/...`, and `golangci-lint run ./...` all clean.

pkg/configuration/manager.go
almaslennikov: Re-reviewed with the intended squish clarified. I checked the MFT implementation: `--force` creates the commander with the maximum port range and uses `SkipChecksLevel::RespectBlocklist`, bypassing capability/dependency/rule checks except explicitly blocklisted TLVs. That supports staging hidden post-breakout params for the same next boot. In non-force mode only exposed `NextBootConfig` keys are sent and normal MFT validation remains active, so visible post params are intentionally squished while hidden ones defer. With that contract, this is not a blocker; withdrawing the finding.

pkg/configuration/manager.go
almaslennikov: Confirmed fixed in `1f7dca3`: combined validation now distinguishes next-boot state from current state, and the new regression test covers update=false/reboot=true for an already-staged Spectrum-X value.

pkg/configuration/manager.go
almaslennikov: Confirmed fixed in `1f7dca3`: after `set_system_conf`, apply re-queries next-boot state before deciding which overrides to send. A focused test with different pre/post query values would still be useful, but the implementation resolves the stale-snapshot bug.

pkg/configuration/manager.go
almaslennikov: Confirmed fixed in `1f7dca3`: the force path now iterates every queried PCI target. A two-PF regression test would be useful, but the implementation resolves the PF0-only behavior.

pkg/configuration/manager.go
almaslennikov: The exact-name coverage issue is addressed in `1f7dca3` for small ranges. Re-review found separate problems in how ranges are expanded and value-validated; I am posting those against the new code.

pkg/nvconfig/utils.go
almaslennikov: The compile-time interface break is fixed in `1f7dca3` by splitting out `SystemConfValidator`. Re-review found that the missing-extension fallback silently disables Network Bay behavior; I am posting that separately against the new code.

pkg/nvconfig/utils.go
almaslennikov: Confirmed fixed in `1f7dca3`: partial parsed rows plus a nonzero command error are rejected when no trailing `Result:` line was seen. The exact caller-level regression test is still missing, but the implementation handles the reported failure mode.

pkg/configuration/manager.go
```diff
@@ -104,13 +102,52 @@ func (h configurationManager) ValidateDeviceNvSpec(ctx context.Context, device *
...
+	if configUpdateNeeded {
+		log.Log.Info("nv config not yet applied to next boot", "device", device.Name)
+	}
+	if len(unsupportedParams) > 0 {
+		log.Log.Info("some nv config params are unsupported on this device and will be skipped", "device", device.Name, "params", unsupportedParams)
+	}
+
+	// 5. system_conf coverage: a mismatched profile param not covered (range-aware) by the override config
+	//    means the baseline itself drifted and set_system_conf must be re-applied (reboot-required).
+	if systemConfDrifted(overrides, systemConfMismatched) {
```
almaslennikov, line 134: [P1] Validate expanded range values before treating drift as covered

`systemConfDrifted` expands a range for name coverage, but `validateTemplateParamsApplied` still looks up the literal key such as `MODULE_SPLIT_M0[0..3]`. `QueryNvConfig` exposes the individual keys, so the literal range is classified as unsupported and never value-checked. The coverage check then suppresses all individual MISMATCH rows solely because the range name exists. Actual indices can remain at the wrong values while validation returns update=false/reboot=false; the new test at line 1354 codifies exactly that false convergence with an empty query. Please expand the desired range/value for validation and non-force apply as well, or reject range syntax rather than using it to suppress drift.

pkg/configuration/manager.go
```diff
@@ -701,21 +658,55 @@ func getRawNvConfigParams(device *v1alpha1.NicDevice) map[string]string {
...
+			set[key] = struct{}{}
+			continue
+		}
+		lo, errLo := strconv.Atoi(m[2])
+		hi, errHi := strconv.Atoi(m[3])
+		if errLo != nil || errHi != nil || lo > hi {
+			set[key] = struct{}{}
+			continue
+		}
+		for i := lo; i <= hi; i++ {
```
almaslennikov: [P1] Do not materialize an unbounded user-controlled range

`NvConfigParam.Name` has no bound on the indices, so a value like `X[0..1000000000]` makes every reconcile spend huge CPU/memory building this set. Worse, `X[9223372036854775807..9223372036854775807]` is accepted on a 64-bit host; after the first iteration `i++` overflows, `i <= hi` stays true, and the daemon effectively loops forever. Compare the finite mismatch names against parsed intervals without expanding them, or impose a small validated bound and use an overflow-safe loop.

pkg/configuration/manager.go
```diff
@@ -561,110 +565,63 @@ func (h configurationManager) ResetNicFirmware(ctx context.Context, device *v1al
...
-
-// applySystemConf applies the requested Network Bay system_conf for the device's ASIC if it is not
-// already applied. It is a no-op (ApplyStatusNothingToDo) for non-Network-Bay devices. set_system_conf
-// is persistent and reboot-required, so a successful apply reports RebootRequired.
-func (h configurationManager) applySystemConf(ctx context.Context, device *v1alpha1.NicDevice, options *types.ConfigurationOptions) (*types.ConfigurationApplyResult, error) {
-	if !hasNetworkBaySpec(device) {
-		return &types.ConfigurationApplyResult{Status: types.ApplyStatusNothingToDo}, nil
+	validator, ok := h.nvConfigUtils.(nvconfig.SystemConfValidator)
+	if !ok {
+		log.Log.Info("nvConfigUtils does not implement SystemConfValidator; skipping Network Bay system_conf management", "device", device.Name)
```
almaslennikov: [P2] Do not report convergence when the optional validator is absent

Separating `SystemConfValidator` fixes source compatibility, but this fallback silently disables the requested Network Bay configuration. A custom `NVConfigUtils` that still supplies the existing `ValidateSystemConf` and `SetSystemConf` methods now returns an empty mismatch list, so validate/apply report convergence and never call `SetSystemConf`. Please use the boolean validator conservatively where possible, or return an explicit unsupported-capability error for a Network Bay template; a silent no-op is not a safe compatibility fallback.

pkg/configuration/manager.go
```diff
@@ -561,110 +565,63 @@ func (h configurationManager) ResetNicFirmware(ctx context.Context, device *v1al
...
-	if matches {
-		return &types.ConfigurationApplyResult{Status: types.ApplyStatusNothingToDo}, nil
+		return nil, err
 	}
 
-	log.Log.Info("applying Network Bay system_conf", "device", device.Name, "conf", conf, "asic", asic)
-	if err := h.nvConfigUtils.SetSystemConf(ctx, pci, conf, asic, options.Force); err != nil {
-		log.Log.Error(err, "failed to apply system_conf", "device", device.Name)
-		return &types.ConfigurationApplyResult{Status: types.ApplyStatusFailed}, err
+	var mismatched []string
```
almaslennikov, line 638: [P2] Fail closed when the result says mismatch but no rows were recognized

This caller ignores `result.Matches`. A complete output containing `Result: Device configuration does NOT match...` but no recognized MISMATCH rows yields `Matches=false` and an empty entry list, which this code treats exactly like a matching profile. The new partial-output guard does not cover this because the Result line is present. Return an error (or conservatively request baseline apply) when `!result.Matches` and no mismatch names were parsed, and add a result-only mismatch test.

pkg/nvconfig/mocks/NVConfigUtils.go
```diff
@@ -143,6 +145,36 @@ func (_m *NVConfigUtils) ValidateSystemConf(ctx context.Context, pciAddr string,
 	return r0, r1
 }
 
+// ValidateSystemConfDetailed provides a mock function with given fields: ctx, pciAddr, conf, asic
```
almaslennikov: [P2] Keep the optional extension out of the base generated mock

This file says it is generated for `NVConfigUtils`, but `ValidateSystemConfDetailed` is no longer part of that interface. A normal mockery regeneration will delete this method, the mock will stop satisfying `SystemConfValidator`, and Network Bay tests will silently take the skip fallback. Generate a mock for a composite interface, or put the extension method in a separate non-generated file; compile-time assertions for both interfaces would catch drift.

almaslennikov (commented):
Re-review of `1f7dca3`: the staged-reboot, stale-query, PF0-only force, interface-shape, and partial-output findings are fixed. I also verified the intended squished Spectrum-X behavior against MFT and withdrew that objection. The remaining inline findings are new issues in range handling and compatibility fallback.

pkg/configuration/manager.go
almaslennikov: Fixed. Ranges are now expanded into individual key+value entries in `buildOverrideParams` (e.g. `MODULE_SPLIT_M0[0..3]=1` → `[0]=1`..`[3]=1`), so each index is value-validated by `validateTemplateParamsApplied` against next boot and applied as a concrete key — not just used for name coverage. Coverage is now plain exact-name membership over the expanded map. Replaced the false-convergence test with one where `[2]` is wrong (asserts update=true) plus a converged-when-all-match case.

pkg/configuration/manager.go
almaslennikov: Fixed. Expansion is bounded by `maxRangeSpan` (256) — a wider range returns an error instead of materializing — and the expansion loop breaks on `i == hi` before incrementing, so a single-element range at `math.MaxInt` can't overflow into an infinite loop. `hi-lo` is overflow-safe for the non-negative bounds the regex yields. Added a test that `X[0..1000000000]` errors with "exceeds limit".

pkg/configuration/manager.go
almaslennikov: Fixed — now fails closed. A Network Bay template with an nvConfigUtils that doesn't implement `SystemConfValidator` returns an explicit error ("requests Network Bay system_conf but nvConfigUtils does not implement SystemConfValidator") rather than an empty mismatch list, so validate/apply no longer silently report convergence and skip set_system_conf. Test added (`baseOnlyNVConfig` stub).

pkg/configuration/manager.go
almaslennikov: Fixed. `systemConfMismatchedParams` now checks `result.Matches`: a non-matching result with zero parsed MISMATCH rows returns an error ("reports a mismatch but no mismatched params were parsed") so the reconcile retries instead of treating it as a match. Test added.

pkg/nvconfig/mocks/NVConfigUtils.go
almaslennikov: Fixed. `ValidateSystemConfDetailed` moved out of the generated `NVConfigUtils.go` into a hand-maintained `pkg/nvconfig/mocks/systemconfvalidator.go` with compile-time assertions (`var _ nvconfig.NVConfigUtils` and `var _ nvconfig.SystemConfValidator = (*NVConfigUtils)(nil)`). A mockery regen of NVConfigUtils.go no longer drops it, and the assertions fail the build if the mock stops satisfying either interface.

almaslennikov:
Round 2 addressed (force-pushed `3708ee9`). Thanks for confirming the first batch and for withdrawing the squish objection after checking the MFT `--force` path. The 5 new findings:

- **#134 (P1) range value-check** — ranges are now expanded into per-index key+value entries in `buildOverrideParams`, so each index is value-validated and applied against the concrete keys, not just name-covered. Replaced the false-convergence test with a wrong-index case (update=true) + an all-match case.
- **#688 (P1) unbounded range** — capped at 256 indices (error beyond), and the loop breaks on `i == hi` before incrementing, so `X[max..max]` can't overflow into an infinite loop. Rejection test added.
- **#604 (P2) missing validator** — fails closed with an explicit unsupported-capability error for a Network Bay template instead of silently skipping set_system_conf.
- **#618 (P2) result-only mismatch** — `!Matches` with no parsed MISMATCH rows now returns an error (fail closed), not a silent match.
- **mocks#148 (P2)** — `ValidateSystemConfDetailed` moved to a hand-maintained `systemconfvalidator.go` with `var _` compile assertions for both interfaces, so a mockery regen can't silently drop it.

`go build`, `go test ./pkg/configuration/... ./pkg/nvconfig/... ./internal/controller/...`, and `golangci-lint run ./...` all clean.

pkg/configuration/manager.go
```diff
@@ -192,152 +206,150 @@ func (h configurationManager) ApplyNVConfiguration(ctx context.Context, device *
...
 
-// applySpectrumXNVConfiguration handles the Spectrum-X NV configuration path (breakout + postBreakout)
-func (h configurationManager) applySpectrumXNVConfiguration(ctx context.Context, device *v1alpha1.NicDevice, options *types.ConfigurationOptions) (*types.ConfigurationApplyResult, error) {
-	pci := device.Status.Ports[0].PCI
+	mergeParams(combined, getRawNvConfigParams(device))
 
-	breakoutParams, err := h.spectrumXConfigManager.GetBreakoutMlxConfig(device)
+	// Expand indexed-range keys into per-index entries so each index is value-validated and applied
+	// against the concrete keys QueryNvConfig / validate_system_conf expose, not the literal range key.
+	expanded, err := expandRangeParams(combined)
```
almaslennikov: **P1: preserve layer priority across range expansion.** `combined` can contain a lower-priority concrete key such as `MODULE_SPLIT_M0[2]=0xff` and the higher-priority raw override `MODULE_SPLIT_M0[0..3]=1`. Both normalize to `[2]`, but `expandRangeParams` iterates a Go map, so whichever entry is visited last wins nondeterministically. That violates the documented `rawNvConfig > Spectrum-X/template` priority and can make validation/apply stage the wrong value. Please expand each layer before merging in priority order (or otherwise make collisions priority-aware), with tests for range-over-specific and specific-over-range.

docs/nic-configuration-library.md
```diff
@@ -164,11 +164,10 @@ func NewConfigurationManager(
...
-6. **Optional reset** — `mlxfwreset` unless `ConfigurationOptions.SkipReset=true`
+2. **Network Bay `set_system_conf` (baseline, applied first)** — for ConnectX-9 Network Bay devices (`template.networkBay` set and `status.networkBay` detected), the per-ASIC `set_system_conf <conf>[<asic>]` is applied **before** the regular / Spectrum-X params so those layer on top of it (override priority: `rawNvConfig` > Spectrum-X > system_conf). Drift is detected per-param via the `nvconfig.SystemConfValidator` extension (`ValidateSystemConfDetailed()`), which ignores MISMATCH rows for params owned by a higher-priority layer (range keys like `MODULE_SPLIT_M0[0..3]` are expanded so they cover the individual mismatch rows); a change requires a reboot. Skipped for devices with `ResetToDefault`, or when the injected `NVConfigUtils` does not implement `SystemConfValidator`
```
almaslennikov, line 167: **Non-blocking documentation follow-up:** the latest code now returns an error when a Network Bay template is used with an `NVConfigUtils` that lacks `SystemConfValidator`; it no longer skips or falls back. This paragraph and the `SystemConfValidator` comments still describe the old skip/fallback behavior, so they should be updated when convenient.

pkg/configuration/manager.go
almaslennikov: Fixed — range expansion is now priority-aware. Each layer is expanded and merged in increasing priority order (Spectrum-X breakout/postBreakout → template → rawNvConfig last) via `mergeExpandedRanges`, instead of expanding the already-merged map. So a rawNvConfig range like `MODULE_SPLIT_M0[0..3]` deterministically overrides a lower-priority concrete `MODULE_SPLIT_M0[2]`, and a rawNvConfig concrete key overrides a lower-priority range, regardless of Go map iteration order. Added both tests (range-over-specific and specific-over-range) asserting the exact applied per-index values.

pkg/configuration/manager.go
```diff
@@ -701,21 +674,66 @@ func getRawNvConfigParams(device *v1alpha1.NicDevice) map[string]string {
...
+		lo, errLo := strconv.Atoi(m[2])
+		hi, errHi := strconv.Atoi(m[3])
+		if errLo != nil || errHi != nil || hi < lo {
+			// Not a usable numeric range (e.g. indices overflow int) — keep the literal key.
+			out[key] = val
+			continue
+		}
+		// hi-lo is overflow-safe for non-negative lo<=hi; the loop breaks on equality so it never
+		// increments past hi (which would overflow when hi == math.MaxInt).
+		if hi-lo >= maxRangeSpan {
```
almaslennikov, line 734: Fixed — added an aggregate cap. `mergeExpandedRanges` tracks the running total across all layers and errors at `maxExpandedParams` (4096) as entries are added, so an abusive input is rejected before it is fully materialized (not just the per-range `maxRangeSpan` of 256). Added a test that 20×256-entry ranges error with "refusing to materialize". I kept it as a total-materialization guard rather than non-materializing interval comparison, since apply still needs the concrete per-index keys to send to mlxconfig anyway; the cap bounds peak memory well under the 128 MiB limit.

almaslennikov:
Round 3 addressed (force-pushed `b32248f`):

- **range expansion priority (P1)** — ranges are now expanded *per layer* and merged in increasing priority order (Spectrum-X → template → rawNvConfig last) via `mergeExpandedRanges`, so a higher-priority range or concrete key deterministically wins over a lower-priority collision regardless of map iteration order. Tests added for both range-over-specific and specific-over-range, asserting exact per-index applied values.
- **aggregate expansion cap (P1)** — `mergeExpandedRanges` now tracks the running total across all layers and rejects at `maxExpandedParams` (4096) as entries are added, before full materialization — so a CR full of in-bounds ranges can't OOM the daemon. Test added (20×256 ranges → "refusing to materialize").

`go build`, `go test ./pkg/configuration/... ./pkg/nvconfig/...`, and `golangci-lint run ./...` clean; `internal/controller` passes on rerun (the two failures were the pre-existing envtest maintenance-suite flake, unrelated to this change).

pkg/configuration/manager.go
almaslennikov: Confirmed on `b32248f`: expanding each layer before the ordered merge fixes the cross-layer raw-over-Spectrum-X collision, and both collision directions are covered. Thanks.

pkg/configuration/manager.go
```diff
@@ -701,21 +684,79 @@ func getRawNvConfigParams(device *v1alpha1.NicDevice) map[string]string {
...
+func mergeExpandedRanges(dst, src map[string]string) error {
+	add := func(key, val string) error {
+		dst[key] = val
+		if len(dst) > maxExpandedParams {
+			return fmt.Errorf("expanded nv config exceeds %d params; refusing to materialize", maxExpandedParams)
+		}
+		return nil
+	}
+
+	for key, val := range src {
```
almaslennikov, line 715: One functional overlap case remains within a single layer. `rawNvConfig` may contain both `M[0..3]=1` and an explicit exception `M[2]=5`; both are accepted by mlxconfig, but because `src` is a Go map this loop randomly decides which value survives for `M[2]`. Please either make same-layer overlaps deterministic (for example, ranges first and exact indices second) or reject them explicitly, with one raw-within-raw test. The new cross-layer tests do not exercise this case.

almaslennikov:
Superseded by #389 — re-opened from a clean branch with a single squashed commit for a tidy history. The code is identical to the final reviewed state here; all Greptile findings from this thread were addressed. Closing this one to move review to #389.

PR #393, opened 2026-07-07
fix: optimize NV configuration apply flow

almaslennikov:
## Summary
- Avoid the initial full `mlxconfig query` in `ApplyNVConfiguration`; use a targeted `LINK_TYPE_P1` capability query and a reset-only query path.
- Apply `set_system_conf` before per-port diffing, then query each port only when a non-force diff is needed.
- Extrapolate port-suffixed params from `NUM_OF_PF`, preserve `WithDefault` semantics for query-visible params only, and improve ApplyNVConfiguration info/debug logging.

## Validation
- `git diff --check upstream/main..deprecate-max-acc-out-read`
- `git diff --check deprecate-max-acc-out-read..optimize-apply-nv-config`
- `go test ./pkg/configuration/... ./pkg/nvconfig/... -v`

### DIFF
--- a/docs/nic-configuration-library.md
+++ b/docs/nic-configuration-library.md
@@ -163,11 +163,11 @@ func NewConfigurationManager(
 
 #### NV Configuration Flow
 
-1. **Query** current NV config via `nvConfigUtils.QueryNvConfig()`
-2. **Network Bay `set_system_conf` (baseline, applied first)** — for ConnectX-9 Network Bay devices (`template.networkBay` set and `status.networkBay` detected), the per-ASIC `set_system_conf <conf>[<asic>]` is applied **before** the regular / Spectrum-X params so those layer on top of it (override priority: `rawNvConfig` > Spectrum-X > system_conf). Drift is detected per-param via `nvconfig.ValidateSystemConf()`, which returns the overall match bit plus the names of the mismatched params; the manager ignores MISMATCH rows for params owned by a higher-priority layer (matched by exact per-index key — `rawNvConfig` index-range syntax like `MODULE_SPLIT_M0[0..3]` is rejected by CRD validation, so keys are always concrete); a change requires a reboot. Skipped for devices with `ResetToDefault`
-3. **Diff** desired vs current parameters. Params not present in the device's next-boot config are unsupported on this device (e.g. hidden because `ADVANCED_PCI_SETTINGS` is off) and are skipped — the operator does **not** auto-manage `ADVANCED_PCI_SETTINGS`; drive it explicitly via `template.rawNvConfig` if needed
-4. **Batch set** via `nvConfigUtils.SetNvConfigParametersBatch()` — single `mlxconfig set` call (`--force` added when `ConfigurationOptions.Force=true`). Apply reports `ApplyStatusPartiallyApplied` when any desired param was skipped as unsupported
-5. **Optional reset** — `mlxfwreset` unless `ConfigurationOptions.SkipReset=true`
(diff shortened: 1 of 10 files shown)


almaslennikov:
Addressed the latest feedback in amended commit `ec31564`:

1. **`--with_default` no-op status**: `SetNvConfigParametersBatch` now returns `types.ApplyStatus` and parses mlxconfig output. A successful `--with_default` command without a `Configurations:` section returns `ApplyStatusNothingToDo`; output with `Configurations:` returns `ApplyStatusSuccess`. `ApplyNVConfiguration` now only marks params applied when the setter reports success. The with-default command also uses `-e --with_default -y`.
2. **Non-deterministic `_P*` extrapolation**: extrapolation now deterministically uses the lowest existing port index as the base value, preserving P1 preference and avoiding map-iteration-dependent results when P1 is absent.
3. **LINK_TYPE_P1 unsupported query handling**: the targeted query now treats mlxconfig exit status 3 as unsupported for this parameter and still recognizes support-message variants, instead of depending only on one exact output phrase.

Validation:
- `GOCACHE=/private/tmp/codex-go-cache go test ./pkg/configuration/... ./pkg/nvconfig/... -v`
- `GOCACHE=/private/tmp/codex-go-cache GOLANGCI_LINT_CACHE=/private/tmp/codex-golangci-cache ./bin/golangci-lint-v2.11.4 run ./pkg/configuration/... ./pkg/nvconfig/...`
- `GOCACHE=/private/tmp/codex-go-cache go build ./...`
- `git diff --check`

Full `go test ./... -count=1` was also run outside the sandbox: all non-e2e packages passed; `test/e2e` failed because no Kubernetes cluster is available (`kubectl` connects to localhost and gets connection refused).

almaslennikov:
Addressed the latest Greptile finding in amended commit `2052146`:

1. **Bare `exit status 3` masked unrelated LINK_TYPE_P1 query failures**: tightened `isUnsupportedLinkTypeParamError` so it only suppresses the targeted query error when the error text includes `LINK_TYPE_P1` and an unsupported-parameter message. A plain `exit status 3` now propagates as an apply failure instead of disabling link type management silently.
2. **Regression coverage**: updated the targeted-query test to assert plain `exit status 3` returns `ApplyStatusFailed`, while the real unsupported output path remains `NothingToDo`.

Validation after the amend:
- `GOCACHE=/private/tmp/codex-go-cache go test ./pkg/configuration/... ./pkg/nvconfig/... -v`
- `GOCACHE=/private/tmp/codex-go-cache GOLANGCI_LINT_CACHE=/private/tmp/codex-golangci-cache ./bin/golangci-lint-v2.11.4 run ./pkg/configuration/... ./pkg/nvconfig/...`
- `git diff --check`

almaslennikov:
Follow-up: force-pushed the actual amended commit containing the Greptile fix: `2a9a37e`.

The final diff tightens `isUnsupportedLinkTypeParamError` to require both `LINK_TYPE_P1` and an unsupported-parameter message before suppressing the targeted query failure. Plain `exit status 3` now propagates as `ApplyStatusFailed`, covered by the updated regression test.

Validation remains:
- `GOCACHE=/private/tmp/codex-go-cache go test ./pkg/configuration/... ./pkg/nvconfig/... -v`
- `GOCACHE=/private/tmp/codex-go-cache GOLANGCI_LINT_CACHE=/private/tmp/codex-golangci-cache ./bin/golangci-lint-v2.11.4 run ./pkg/configuration/... ./pkg/nvconfig/...`
- `git diff --check`

Gal-Zaidman:
overall looks good to me, woud be nice to not require a reboot after an mlxconfigure reset command, and instead continue with the usual flow till the end.

PR #394, opened 2026-07-07
docs: deprecate maxAccOutRead

almaslennikov:
## Summary
- Mark `template.pciPerformanceOptimized.maxAccOutRead` as deprecated/ignored in the API comments and generated CRD descriptions.
- Remove `maxAccOutRead` from samples and examples.
- Update README and API/library docs to direct explicit `MAX_ACC_OUT_READ` management through `rawNvConfig`.

## Validation
- `GOCACHE=/private/tmp/codex-go-cache make manifests`
- `git diff --check upstream/main..deprecate-max-acc-out-read`
- `go test ./pkg/configuration/... ./pkg/nvconfig/... -v`

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,6 @@ spec:
       linkType: Ethernet
       pciPerformanceOptimized:
          enabled: true
-         maxAccOutRead: 44
          maxReadRequest: 4096
       roceOptimized:
          enabled: true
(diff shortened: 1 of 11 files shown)


almaslennikov:
Addressed the Greptile finding in amended commit `d090a72`:

1. **Forbidden `deprecated: true` in CRD schemas**: removed the manual `deprecated: true` entries from all generated CRD copies. The CRDs now match `controller-gen` output from the Go `// Deprecated:` field comment: the deprecation is carried in the generated schema description only.

Validation after the amend:
- `GOCACHE=/private/tmp/codex-go-cache make manifests`
- `git diff --check upstream/main..deprecate-max-acc-out-read`
- `go test ./pkg/configuration/... ./pkg/nvconfig/... -v`

PR #405, opened 2026-07-09
perf: batch DMS get parameter queries

almaslennikov:
## Summary

- Batch `DMSClient.GetParameters` reads by concrete interface: each interface uses one `dmsc get` command with repeated `--path` arguments, while paths without an interface filter use a separate global batch.
- Preserve existing interface, priority, and `HwplbFirstPortOnly` expansion without batching multiple interfaces into the same DMS command.
- Parse batched output by the returned `Path`, independent of response order. DMS's selector-free `values` key is used only to read the value; results remain keyed by the original configured DMS path so selectors such as `slot[id=0]` and `param[id=11]` are preserved.
- Detect differing values for the same parameter within one batch and across interface batches, including when the first value is empty.
- Capture combined stdout/stderr and log raw DMS output at debug level.
- Document the batching behavior and add regression coverage for raw multi-result output, repeated updates, ordering, static selectors, per-interface routing, priorities, HWPLB first-port handling, and error paths.

## Validation

- `go test ./pkg/dms -v -count=1 -timeout 2m -run TestDMS --ginkgo.focus='DMSClient'`
- `go test ./pkg/spectrumx/... -count=1 -timeout 5m`
- `go test ./pkg/configuration/... -count=1 -timeout 5m`
- `go vet ./pkg/dms/...`
- `go build ./...`
- `./bin/golangci-lint-v2.11.4 run`
- `git diff --check upstream/main...HEAD`

## Test Image

- `harbor.mellanox.com/cloud-orchestration-dev/nic-configuration-operator-daemon:batch-dms-get-parameters` (`linux/amd64`)

### DIFF
--- a/docs/nic-configuration-library.md
+++ b/docs/nic-configuration-library.md
@@ -271,6 +271,12 @@ dmsc -a <address> <authParams...> --target 0000:08:00.0 set --update "path:::typ
 - `--tls-key <path>` — client TLS private key
 - `--skip-verify` — skip server certificate verification
 
+#### Batch GetParameters
+
+`GetParameters` expands interface and priority parameters, then groups the resulting paths by concrete interface. Each interface is queried with one `dmsc get` command containing multiple `--path` flags; paths without an interface filter use a separate global batch. Parameters marked `hwplbFirstPortOnly` are added only to the first interface batch.
+
+The response parser matches each returned `Path` to its request, reads the value through DMS's selector-free `values` key, and stores it under the request's original parameter path. This removes generated interface and priority selectors while preserving configured selectors such as `slot[id=0]` and `param[id=11]`. Values returned for the same parameter within a batch or across interface batches must match.
+
 #### Batch SetParameters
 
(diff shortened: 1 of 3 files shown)


almaslennikov:
Updated the PR in amended commit `84e4db3`:

1. **Order-independent parsing**: batched DMS responses are matched by their returned `Path`; no positional assumptions remain.
2. **Correct result keys**: DMS's selector-free `values` key is used only for lookup, while parsed results use the original configured parameter path so static selectors such as `slot[id=0]/param[id=11]` are preserved.
3. **Per-interface correctness**: get requests are batched per concrete interface, with mismatch detection within a batch and across interface batches; `HwplbFirstPortOnly` remains first-port-only.
4. **Coverage and lint**: added raw-output, ordering, repeated-update, static-selector, empty-value, priority, interface, and HWPLB tests, and fixed the preallocation lint findings.

Validation passed: focused DMS tests, Spectrum-X tests, configuration tests, `go vet ./pkg/dms/...`, `go build ./...`, and the pinned golangci-lint with zero issues.

Gal-Zaidman:
looks good to me

PR #407, opened 2026-07-11
Skip device on discovery errors when enabled

almaslennikov:
## Summary
- add SKIP_DEVICE_ON_DISCOVERY_ERROR support for per-device discovery failures
- skip the affected physical NIC when firmware/PSID lookup or BlueField operation-mode lookup fails and the env var is true
- preserve existing NicDevice CRs for skipped physical PCI keys to avoid transient delete/recreate churn
- document the feature flag in README and chart values

## Testing
- GOCACHE=/private/tmp/nco-skip-discovery-go-cache go test ./pkg/devicediscovery/... -v
- GOCACHE=/private/tmp/nco-skip-discovery-go-cache go test ./internal/controller -v --ginkgo.focus='DeviceDiscoveryController'
- GOCACHE=/private/tmp/nco-skip-discovery-go-cache go build ./...
- GOCACHE=/private/tmp/nco-skip-discovery-go-cache GOLANGCI_LINT_CACHE=/private/tmp/nco-skip-discovery-golangci-cache make lint

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,3 +406,4 @@ Supported flags:
 
 * `SKIP_VM_CHECK`=`true`: pass `--skip_vm_check` to `mlxfwreset` resetting NIC firmware while running on VMs.
 
+* `SKIP_DEVICE_ON_DISCOVERY_ERROR`=`true`: skip a physical NIC during device discovery when firmware/PSID lookup or BlueField operation-mode lookup fails. When unset, these errors fail discovery so the daemon retries. Existing `NicDevice` CRs are preserved for skipped devices to avoid delete/recreate churn during transient per-device discovery failures.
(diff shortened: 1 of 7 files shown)


almaslennikov:
Addressed the Greptile findings and force-pushed amended commit `55b11b3`:

1. **Physical NIC partially discovered**: discovery now tracks skipped devices by collapsed physical PCI key and removes any already-assembled status for that key, so a later PCI function cannot create a partial `NicDevice`.
2. **BlueField ports can disappear**: the same physical-key skip handling is used for BlueField operation-mode lookup failures, preventing truncated port lists and mixed metadata from later functions.
3. **Transient failure deletes CRs**: `DeviceDiscoveryController` now consumes skipped-device reports when available and preserves existing `NicDevice` CRs for skipped physical PCI keys instead of treating them as removed hardware.

Validation: `go test ./pkg/devicediscovery/... -v`, `go test ./internal/controller -v --ginkgo.focus='DeviceDiscoveryController'`, `go build ./...`, and `make lint` are clean locally.

PR #408, opened 2026-07-13
Backport skip device on discovery errors to 26.4

almaslennikov:
## Summary
- Backport #407 to `network-operator-26.4.x` with `git cherry-pick -x`
- Add `SKIP_DEVICE_ON_DISCOVERY_ERROR=true` handling for per-device firmware/PSID and BlueField operation-mode discovery failures
- Preserve existing `NicDevice` CRs when a physical device was skipped during discovery

## Verification
- `go test ./pkg/devicediscovery/... -v`
- `go test -c ./internal/controller -o /tmp/nco-controller-tests-backport.test`
- `go build ./...`

`go test ./internal/controller/... -v` could not run locally because envtest binaries are missing at `../../bin/k8s/1.31.0-darwin-arm64/etcd`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,3 +358,5 @@ The NIC Configuration Daemon itself relies on the `network.nvidia.com/operator.m
 Feature flags can be enabled via environment variables in the helm chart or NVIDIA Network Operator's NicClusterPolicy.
 Supported flags:
 * `FW_RESET_AFTER_CONFIG_UPDATE`=`true`: explicitely reset the NIC's Firmware before the reboot and after updating its non-volatile configuration. Might be required on DGX servers where configuration update is not successfully applied after the warm reboot.
+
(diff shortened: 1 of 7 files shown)


almaslennikov:
Addressed the Greptile finding and force-pushed amended commit `da5d8cb`.

1. **VPD failures bypass flag**: the original PR feedback was about multi-function partial discovery and CR preservation, which was handled with physical-PCI skip tracking plus controller preservation. This backport-specific finding was narrower: VPD failures were being reported through `SkippedDevices()` even when `SKIP_DEVICE_ON_DISCOVERY_ERROR` was unset. I changed the VPD path so it only records a skipped physical PCI key when the flag is enabled; default deployments keep the release branch's existing behavior.
2. **Regression coverage**: added assertions that default VPD failures do not populate `SkippedDevices()`, and that enabled VPD failures do populate it.

Validation: `go test ./pkg/devicediscovery/... -v`, `go test -c ./internal/controller -o /tmp/nco-controller-tests-backport.test`, `go build ./...`, and `git diff --check` are clean locally. The full controller envtest suite still requires the missing local envtest binaries.

PR #410, opened 2026-07-13
Guard VPD skip reporting behind discovery flag

almaslennikov:
## Summary
- keep VPD discovery failures out of `SkippedDevices()` unless `SKIP_DEVICE_ON_DISCOVERY_ERROR=true`
- preserve existing enabled behavior where VPD failures are reported as skipped devices when the flag is set
- add regression coverage for both default and enabled VPD failure behavior

## Testing
- `go test ./pkg/devicediscovery/... -v`
- `go build ./...`
- `git diff --check`

Note: this follows up on the release backport review finding so `main` and the backport keep the same opt-in semantics.

### DIFF
--- a/internal/controller/devicediscovery_controller.go
+++ b/internal/controller/devicediscovery_controller.go
@@ -120,9 +120,16 @@ func (d *DeviceDiscoveryController) reconcile(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	skippedDevices := map[string]error{}
+	incompleteDevices := map[string]error{}
+	if incompleteDeviceReporter, ok := d.deviceDiscovery.(devicediscovery.IncompleteDeviceReporter); ok {
+		for pciKey, err := range incompleteDeviceReporter.IncompleteDevices() {
+			incompleteDevices[pciKey] = err
+		}
+	}
 	if skippedDeviceReporter, ok := d.deviceDiscovery.(devicediscovery.SkippedDeviceReporter); ok {
-		skippedDevices = skippedDeviceReporter.SkippedDevices()
+		for pciKey, err := range skippedDeviceReporter.SkippedDevices() {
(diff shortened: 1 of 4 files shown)


almaslennikov:
Addressed Greptile's VPD cleanup finding and force-pushed amended commit `b6154fd`:

1. **Partial Device Survives VPD Failure**: split partial-status cleanup from skip reporting. A VPD failure now always removes any partially assembled physical device from `statuses`, so discovery cannot return a truncated multi-port `NicDevice`.
2. **Skipped Device Becomes Deletable**: kept CR preservation opt-in. `SkippedDevices()` is still populated for VPD failures only when `SKIP_DEVICE_ON_DISCOVERY_ERROR=true`, matching the flag semantics while preserving the cleanup behavior above.

Validation: `go test ./pkg/devicediscovery/... -v`, `go build ./...`, and `git diff --check` are clean locally.

almaslennikov:
Addressed the new Greptile P1 and force-pushed amended commit `455acc6`:

1. **CR deletion remains**: split discovery omission reporting from reconciliation preservation. `SkippedDevices()` stays controlled by `SKIP_DEVICE_ON_DISCOVERY_ERROR`, while the new `IncompleteDevices()` signal tells the controller to preserve existing CRs for devices omitted because discovery could not build complete status.
2. **Partial/truncated devices**: VPD failures now always mark the physical PCI key incomplete and drop any partially assembled status, so multi-port NICs cannot be reconciled with only the successfully discovered ports.
3. **Controller behavior**: the device discovery controller now preserves CRs for `IncompleteDevices()` and still supports the previous `SkippedDevices()` signal for compatibility.

Validation: `go test ./pkg/devicediscovery/... -v`, `go test -c ./internal/controller -o /tmp/nco-controller-tests-main-followup.test`, `go build ./...`, and `git diff --check` are clean locally.

PR #430, opened 2026-08-05
fix: fail runtime validation when link has no carrier

almaslennikov:
## Summary

- detect the Linux `NO-CARRIER` state before checking runtime configuration values
- return an interface- and PCI-specific validation error so runtime apply stops cleanly
- surface the error through the existing `ConfigUpdateInProgress=False` condition with reason `RuntimeConfigUpdateFailed`
- document the behavior and cover carrier detection, validation failure, and status propagation

## Testing

- `make test`
- `go build ./...`
- `make lint`


almaslennikov:
Updated the test coverage in amended commit `4fa01e7`:

1. **Preserved the generic failure regression:** restored `Should not release maintenance if runtime config failed to apply` unchanged with its original generic error.
2. **Kept carrier coverage additive:** the `NO-CARRIER` status assertion remains a separate test case.

Verified with the two focused envtest specs, `go build ./...`, and `make lint` (0 issues).

PR #431, opened 2026-08-07
feat: apply NVConfig through dms-cli

almaslennikov:
## Summary

- add a package-level `pkg/dmscli` API for the agentless `/nvidia/nvconfig/apply` action, with no client lifecycle or stored runtime state
- reuse the repository-standard `k8s.io/utils/exec.Interface` for context-aware command execution and `FakeExec` testing
- model raw scalar values, typed XPath operations, port fanout, `with-default`, and `force` in the reusable request type
- route the existing `SetNvConfigParametersBatch` seam through `dms-cli` while preserving configuration calculation, per-PF batching, status, and reboot behavior
- omit `ports` and `typed` from the current raw-only action payload
- add `/opt/mellanox/doca/services/dms` to the daemon runtime `PATH`, matching the standard `doca-dms` package layout while keeping the Go package runtime-agnostic
- log the exact `dms-cli` argv and combined stdout/stderr, per-target changed/unchanged/unsupported diff, and structured DMS apply result
- use only the canonical `pci/<BDF>` action target; `FwctlDevice` metadata is neither passed nor reported as the apply target
- keep NVConfig query/reset/system-conf, runtime configuration, Spectrum-X planning, and the existing `dmsc`/`dmsd` paths unchanged in this phase

The public `NVConfigUtils` signature is unchanged, so this is backward compatible for downstream Go consumers.

## Command shape

```text
dms-cli --json -t pci/<BDF> --input <payload-json> /nvidia/nvconfig/apply
```

The command log contains the executable and complete argv as a structured array, including the exact JSON payload, followed by the combined command output. The configuration manager logs why each desired parameter was applied, skipped as already staged, or skipped as unsupported.

## Compatibility

This path requires a `doca-dms` build containing DMS source commit `39e47e6209` (`/nvidia/nvconfig/apply`), first present on the 3.5 line in build `3.5.0069-1`. There is intentionally no direct-`mlxconfig` fallback for this batch apply operation.

## Verification

- `go test ./pkg/dmscli/... ./pkg/nvconfig/... ./pkg/configuration/... -v -timeout 10m`
- `KUBEBUILDER_ASSETS=<absolute 1.31.0 assets path> go test $(go list ./... | rg -v /e2e) -timeout 15m`
- `go build ./...`
- `go vet ./...`
- `make lint`
- `git diff --check upstream/main`

All passed locally. The cluster-dependent e2e suite remains outside the repository unit-test target and was not run without a configured cluster.

### DIFF
--- a/Dockerfile.nic-configuration-daemon
+++ b/Dockerfile.nic-configuration-daemon
@@ -30,6 +30,8 @@ FROM ${BASE_IMAGE_DOCA_FULL_RT_HOST:-nvcr.io/nvstaging/doca/doca:full-rt-3.5.0-h
 
 ARG TARGETARCH
 
+ENV PATH="/opt/mellanox/doca/services/dms:${PATH}"
+
 ARG PACKAGES="dpkg-dev libusb-1.0-0 ipmitool rshim curl systemd-sysv mstflint"
 
 # enable deb-src repos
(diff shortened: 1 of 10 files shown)


almaslennikov:
Addressed the API-structure feedback in amended commit `e8a827a`:

1. **Reuse the existing executor.** Removed the package-local `Runner`/`osRunner` abstraction and now accept the repository-standard `k8s.io/utils/exec.Interface`, retaining `CommandContext`, `CombinedOutput`, V(2) logging, and `FakeExec` testability.
2. **Remove unnecessary runtime objects.** Deleted `dmscli.Client`, its options, and the stored NVConfig applier; `dmscli.ApplyNVConfig` is now a package-level function called with `nvConfigUtils`' existing executor.
3. **Remove structural noise.** Deleted `pkg/dmscli/doc.go`; the current raw-only call omits `Ports` and `Typed` from its request literal and JSON payload. Those request fields remain only for the later typed XPath/doSPCX phase.

Verification is clean: focused package suites, full non-e2e suite with envtest assets, `go build ./...`, `go vet ./...`, `make lint`, and `git diff --check upstream/main`.

almaslennikov:
Image preflight exposed and fixed one runtime integration issue in amended commit `453d995`:

1. **Make the packaged executable discoverable.** `doca-dms 3.5.0079-1` installs `dms-cli` at `/opt/mellanox/doca/services/dms/dms-cli`, but that directory is not on the base image's default `PATH`. The daemon Dockerfile now adds the standard DMS directory to `PATH`, so the package-level API can continue invoking `dms-cli` by name and remain reusable outside this image layout.
2. **Document the contract.** The design and library docs now state that callers must provide `dms-cli` on `PATH` and identify the daemon image's standard DOCA path.

Focused suites, `go build ./...`, `make lint`, and `git diff --check` are clean. The requested DOCA 3.5.0079 image build will use this amended commit.

PR #463, opened 2026-09-01
fix: apply Spectrum-X runtime parameters separately

almaslennikov:
## Summary

- keep Spectrum-X DMS runtime reads batched
- apply each declared runtime parameter separately in profile order
- route DMS and mlxreg parameters through a shared setParamWithRetry path with up to 10 total attempts and a one-second interval
- prefer each port fwctl device for mlxreg sets, falling back to its PCI address
- preserve existing IgnoreError behavior for DMS parameters

## Testing

- go test ./pkg/spectrumx/... -count=1
- go build ./...
- make lint

### DIFF
--- a/docs/nic-configuration-library.md
+++ b/docs/nic-configuration-library.md
@@ -496,7 +496,7 @@ type MlxRegField struct {
 }
 ```
 
-A parameter with `MlxConfig` set is applied via `nvconfig.SetNvConfigParametersBatch()`. Runtime profile parameters with `DMSPath` set are applied via `dms.DMSClient.SetParameters()`. Runtime profile parameters with `MlxReg` set are applied with `mlxreg`; they preserve profile order and split surrounding DMS parameters into separate batches.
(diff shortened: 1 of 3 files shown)


almaslennikov:
Updated in amended commit dfd1793:

1. **Unified runtime retry path.** setParamWithRetry now accepts a ConfigurationParameter and retries both singleton DMS sets and mlxreg sets for up to 10 total attempts.
2. **Fwctl-first mlxreg targeting.** mlxreg set operations use the discovered FwctlDevice for each port and fall back to PCI when fwctl is unavailable.
3. **Expanded unit coverage.** Added mlxreg recovery/exhaustion tests, continuation after recovery, fwctl preference, and PCI fallback coverage while retaining batched-read and DMS retry tests.

Verification is clean: go test ./pkg/spectrumx/... -count=1, go build ./..., and make lint.

heyvister1 (approved):

Gal-Zaidman:
looks good to me

PR #485, opened 2026-09-14
fix: read PCI VPD through sysfs

almaslennikov:
## What this fixes

Fix for: https://redmine.mellanox.com/issues/5269725

NIC discovery currently shells out to `mlxvpd`, then falls back to `mstvpd`. Under Secure Boot/lockdown, those external tools can be prevented from using their direct PCI access path even though the kernel can read the device VPD. Both commands then fail and the discovery pass omits the physical NIC.

## What changed

- Read `/sys/bus/pci/devices/<BDF>/vpd` directly instead of executing either VPD tool.
- Parse the standard binary PCI VPD Identifier String and read-only data resources in pure Go.
- Extract the existing NCO fields: product/model Identifier String, `PN`, and `SN`.
- Bounds-check every resource header, resource payload, keyword header, and keyword payload before slicing, with contextual errors for unreadable, truncated, malformed, or incomplete VPD.
- Keep `PN` and `SN` required and the model/identifier optional, preserving the existing discovery contract.
- Read the requested BDF's sysfs attribute unchanged. Linux PCI core can transparently route a non-zero function through function 0 when those functions share VPD storage.

This uses the kernel-mediated sysfs interface, so Secure Boot policy no longer depends on VPD user-space utilities. The parser was implemented from the PCI VPD resource format; no pciutils source was copied and no dependency was added.

## Validation

- `go test ./pkg/devicediscovery/... -count=1`
- `go build ./...`
- `make test` (Kubernetes 1.31 envtest; non-E2E repository suite)
- `make lint` (`golangci-lint` v2.11.4, 0 issues)


almaslennikov:
Tested on a real cluster

PR #494, opened 2026-09-16
fix: require no carrier on all NIC ports

almaslennikov:
## Summary

- fail the NO-CARRIER runtime validation gate only when every discovered interface on the NIC lacks carrier
- allow runtime configuration to continue when at least one plane is connected
- preserve link-state query errors and the existing RuntimeConfigUpdateFailed status path
- update focused regression coverage and user-facing behavior documentation

This is the network-operator-26.7.x counterpart of #493 and adjusts the behavior introduced by #476 for phased multi-plane deployments.

## Verification

- go test ./pkg/configuration/... -count=1 -timeout 10m
- go build ./...
- make lint
- git diff --check


almaslennikov:
Amended the implementation in commit 137f204 as requested:\n\n1. Carrier validation now returns immediately when it finds the first interface without NO-CARRIER.\n2. If every checked interface has NO-CARRIER, it returns one NIC-level error after the loop.\n3. Ports without a discovered interface retain the existing later validation path and its specific missing-interface error.\n4. The mixed-state regression now puts NO-CARRIER on the first port, proving iteration continues until the second port with carrier is found.\n\nVerified with go test ./pkg/configuration/... -count=1 -timeout 10m, package-scoped golangci-lint, and git diff --check.

almaslennikov:
Correction: the early-return implementation is now actually folded into amended commit 3745ddb and force-pushed. The branch is clean, and the committed source now returns nil immediately on the first false HasNoCarrier result; the all-NO-CARRIER error is emitted after the loop. The same focused tests and package lint passed against this exact tree.

PR #498, opened 2026-09-21
feat: support multiple interface naming templates per node

### LINKED ISSUE #497: RFE: per-fabric-role interface naming (multiple naming groups on one host)
## What would you like to be added?

The ability to apply **different interface name prefixes to different groups of NICs on the same host**, where the grouping and the prefixes are chosen by the operator rather than implied by the API.

The goal is to name interfaces after the **role of the fabric they attach to**, so that a workload can select devices by name without knowing anything about the machine.

Two deployment shapes we care about, both of which are inexpressible today:

**Mixed InfiniBand and Ethernet host**

| Fabric role | Desired name |
| --- | --- |
| InfiniBand compute rails | `ib0`, `ib1`, ... |
| RoCE ports | `roce0`, `roce1`, ... |
| In-band management | `inband0`, ... |

**All-Ethernet host**, using the DGX SuperPOD reference architecture's own terms

| Fabric role | Desired name |
| --- | --- |
| East/West, inter-process communication, RDMA | `ew0`, `ew1`, ... |
| North/South, shared storage access, RDMA | `ns0`, `ns1`, ... |
| Administration / management | `admin0`, ... |

`compute` / `storage` / `mgmt` would work equally well. The point of the request is precisely that the operator should not have to care which taxonomy we pick.

## Why is this needed?

The main benefit is a single portable RDMA configuration.

When the east-west devices, and only those, are named `ew*`, then `NCCL_IB_HCA=ew` selects exactly the right interfaces on every node of every cluster, and can never accidentally select a storage or management port. The same applies to `UCX_NET_DEVICES`, and to a storage client that wants to bind to `ns*` and nothing else.

Kernel names cannot do this. `mlx5_N` numbering follows PCI probe order, so the same logical interface gets a different index on different chassis, and devices belonging to completely different fabrics are interleaved in one flat namespace. That forces per-chassis device lists in workload configuration, which is exactly what we are trying to eliminate.

### Aligning the netdev and RDMA namespaces

A second benefit, which follows from `netDevicePrefix` and `rdmaDevicePrefix` being separate fields, is that the same physical port can be given the **same name in both namespaces**.

By default the two are unrelated and must be correlated by hand through sysfs:

```
netdev  ibp13s0        <->  rdma  mlx5_0
netdev  ens1f0np0      <->  rdma  mlx5_2
```

which is awkward because essentially every tool speaks one namespace or the other. `ip`, `ethtool`, netplan, CNI configuration and most netdev-based exporters use the interface name, while `ibstat`, `rdma`, `mlxlink`, NCCL and UCX use the RDMA device name. Correlating a counter, a log line or an alert with the interface it belongs to means translating between the two every time.

Setting both prefixes to the same value removes the translation step entirely:

```
netdev  ew0            <->  rdma  ew0
```

This part already works today with the current API, and it works well. The limitation is the same as above: it can only be done for the one group a node is allowed to name. Every other NIC on the host keeps a mismatched pair of kernel names. Per-group prefixes would extend a property that is already valuable to the whole machine.

Alignment should stay opt-in per group rather than being implied, since renaming a netdev is not always desirable. A management interface may be referenced by name in netplan, routing or monitoring, so naming only its RDMA device, or only its netdev, is sometimes the right answer. That is the same argument as the `netDevicePrefix` point at the end of this issue.

We do this today with plain udev rules and would like to adopt `NicInterfaceNameTemplate` as the single mechanism. Because it can only name one group per node, we would have to keep a parallel udev mechanism alongside it, which largely defeats the purpose of using the operator.

## Current limitations

Checked against `main`.

To be precise about what does and does not work: naming **one** group is already fine. Setting one NIC per rail with `netDevicePrefix: "ew%rail_id%"` yields `ew0..ew7` today. The blocker is that a node can carry only **one** scheme.

1. **At most one template per node.** Two matching templates is a hard error:

   ```go
   if len(matchingTemplates) > 1 {
       err = fmt.Errorf("multiple NicInterfaceNameTemplates match node %s: %v", r.NodeName, templateNames)
       // Clear the InterfaceNameTemplate spec from all devices on this node
   ```

2. **One prefix pair per template**, `rdmaDevicePrefix` and `netDevicePrefix`.

3. **No device selector.** The spec has no `linkType`, `deviceID`, `vendor` or `psid` field, so a template cannot scope itself to a class of NIC.

4. **The abstraction is rail-oriented.** `railPciAddresses` and `%rail_id%` model a rail-optimized InfiniBand topology. On an all-Ethernet host there are no rails, so the required fields have to be filled in with one NIC per rail purely to obtain a per-device index. It works, but the API is describing something the deployment does not have, and `%rail_id%` ends up meaning "position in the list".

The only separation available today is *omission*: devices absent from `railPciAddresses` get no udev rule. That protects other NICs from being renamed, which we rely on, but it leaves them on unstable kernel names rather than giving them a name of their own.

### Option A: device groups within one template (preferred)

Add an optional list of groups, each with its own prefixes and its own set of devices. Existing top level fields stay valid and behave as the single group case, so this is backward compatible.

A group identifies its devices either by a flat `pciAddresses` list, for topologies with no rail structure, or by the existing `railPciAddresses` plus `pfsPerNic` where rails are meaningful. Indices are per group, so each group numbers from zero.

```yaml
apiVersion: configuration.net.nvidia.com/v1alpha1
kind: NicInterfaceNameTemplate
metadata:
  name: fabric-role-naming
spec:
  nodeSelector:
    node-role.example.com/gpu: ""
  deviceGroups:
    # East/West compute fabric, rail-optimized, so rail semantics are useful
    - name: east-west
      rdmaDevicePrefix: "ew%rail_id%"
      netDevicePrefix: "ew%rail_id%"
      pfsPerNic: 1
      railPciAddresses:
        - ["0000:0d:00.0"]
        - ["0000:35:00.0"]

    # North/South storage fabric, no rail structure, just an ordered list
    - name: north-south
      rdmaDevicePrefix: "ns%index%"
      netDevicePrefix: "ns%index%"
      pciAddresses:
        - "0000:48:00.0"
        - "0000:48:00.1"

    # Administration, netdev only, no RDMA naming
    - name: admin
      netDevicePrefix: "admin%index%"
      pciAddresses:
        - "0000:c1:00.0"
```

`%index%` would be the position within the group. A PCI address appearing in more than one group would be a validation error.

### Option B: allow multiple templates per node when they do not overlap

Keep one prefix pair per template, but permit several templates to match a node as long as their device sets are disjoint, erroring only on genuine overlap. Smaller change, though the effective configuration for a node then has to be reassembled from several objects.

### Option C: a device selector

Add matching on `linkType`, device ID or PSID, so a group can be described rather than enumerated. Attractive in its own right, and it composes with Option A as an alternative to listing PCI addresses. Note it is not sufficient on its own: link type alone cannot separate east-west from north-south when both are Ethernet with RDMA, which is exactly the case in the second table above.

## Two smaller points, related

**The multi-template failure mode is unnecessarily destructive.** When two templates match, the controller does not merely refuse to apply the second, it clears `InterfaceNameTemplate` from *every* device on the node. A configuration mistake therefore removes naming that was previously working, rather than leaving the last good state in place. Erroring without wiping would be safer, independently of this RFE.

**`netDevicePrefix` is required while `rdmaDevicePrefix` is optional.** For an InfiniBand deployment it can be desirable to rename the RDMA device while leaving the IPoIB netdev alone, which is not expressible today. Making `netDevicePrefix` optional too, with validation that at least one of the two is set, would round this out.

## Anything else?

Happy to test on ConnectX-7, ConnectX-8 and BlueField-3 hardware, including hosts carrying several of these fabric roles at once.

almaslennikov:
## Summary

- allow multiple `NicInterfaceNameTemplate` resources to match one node when they select disjoint `NicDevice` objects
- treat a whole `NicDevice` as the ownership boundary and reject overlapping templates without clearing the last valid assignment
- validate the complete proposed netdev/RDMA name set before updating devices or udev rule files
- reconcile templates when device discovery or local node labels change
- document multi-fabric-role, net-only, and RDMA-only naming behavior

## Compatibility

This does not change the CRD or API schema. Existing single-template configurations keep their current behavior. Each disjoint template calculates its indices independently, while conflicting ownership or duplicate generated names is reported and leaves the last valid configuration in place.

## Testing

- `make test` on Linux/amd64 (includes generation, formatting, vet, unit tests, and envtest)
- `make build` on Linux/amd64
- `make lint` on Linux/amd64 (`0 issues`)
- parsed the three-document fabric-role example as YAML

Closes #497


almaslennikov:
## Real-cluster validation

Validated commit `2bc4fdf7bcb1da1d5075653107d99511d1278183` on a two-worker cluster with eight exposed `NicDevice` resources (four per worker).

### Image under test

- Image: `harbor.mellanox.com/cloud-orchestration-dev/nic-configuration-operator-daemon:multi-interface-templates-2bc4fdf-doca360-v2-amd64`
- OCI index digest: `sha256:24c085d05c8dcbc5e9c9d8e4743551a314ec4855de9c43cc604f54f08b2869ff`
- linux/amd64 manifest digest: `sha256:4871c9555bb79f9f70295bb42ec7db2b36540a9552e961ac19641a4672e628af`
- DOCA base: `nvcr.io/nvstaging/doca/doca:full-rt-3.6.0-host-dev` (`sha256:b47a5e2885f688a692c61c5b96eb686f4eee1eeb63063d8ece831515b7ee3d82`)
- Runtime package checks: `doca-dms 3.6.0045-1`, `doca-runtime 3.6.0-029000`; `dmsd` present
- Both daemon pods became Ready with zero restarts and the NicClusterPolicy returned to `ready`.

An initial image smoke-test used `full-rt-3.6.0013-amd64-host-dev`; it was rejected before feature testing because that image does not contain `/opt/mellanox/doca/services/dms/dmsd`. The policy was immediately rolled back. The `v2` image above is the image used for every feature test.

### Test matrix

1. **Multiple disjoint templates on the same nodes — passed**
   - Applied two server-side-dry-run-validated templates matching both workers in addition to the pre-existing CX8 template.
   - Template A owned CX9 NIC `0001:64:00.0` and generated `t497a0`, `t497a1`, `t497ra0`, and `t497ra1`.
   - Template B owned CX9 NIC `0001:d6:00.0` and generated `t497b0`, `t497b1`, `t497rb0`, and `t497rb1`.
   - Both workers converged independently. Each template received zero-based `nicIndex`, `railIndex`, and `planeIndices` scoped to that template.
   - All affected devices reported `InterfaceNameApplied=True` for their current generation.

2. **Coexistence with an existing matching template — passed**
   - Three templates simultaneously matched each worker: the existing two-rail CX8 template plus templates A and B.
   - Existing CX8 `NicDevice` specs, names (`eth_r0_p*`, `eth_r1_p*`, `roce_r0`, `roce_r1`), and conditions were unchanged.

3. **Overlapping ownership of the same PF — passed**
   - Added a third test template selecting the PF already owned by template A.
   - Both nodes emitted actionable `SpecError` events identifying the two templates and the conflicting `NicDevice`.
   - Existing specs and actual interface names were retained.
   - Net and RDMA udev-rule checksums remained byte-identical on both nodes.

4. **Whole-NicDevice ownership across different PFs — passed**
   - Added a template selecting `0001:d6:00.1` while template B selected `0001:d6:00.0` on the same dual-port NIC.
   - Both nodes rejected the split ownership with `SpecError` naming both templates and the common `NicDevice`.
   - No spec or udev-rule change was persisted.

5. **Generated network-name collision — passed**
   - Added a disjoint template whose generated names collided with template B (`t497b0` and `t497b1`).
   - Both nodes rejected it with an actionable `SpecError` containing the duplicate name and both PCI/device identities.
   - Device specs and both udev-rule files remained unchanged.

6. **Independent template deletion — passed**
   - Deleted template A while leaving template B and the original CX8 template present.
   - Only A-owned CX9 specs were cleared and only A rules were removed; template B and the original CX8 state remained intact.

7. **Full cleanup and restoration — passed**
   - Deleted all temporary templates and confirmed only the original template remains with its original spec.
   - Restored the four tested CX9 network and RDMA names to their captured baseline values; all four CX9 specs are empty again.
   - Removed the test-added `InterfaceNameApplied` status conditions from those four CX9 resources so their condition sets match the baseline.
   - Confirmed no `t497` or `issue-497` strings remain in either host's naming rules or live interface names.
   - Both hosts have matching baseline-only rule hashes:
     - net: `a5b46507ef923bb26a9ca4244811b0ebff207e8f7e42ea40efb31bc30e5938a4`
     - RDMA: `a46da42ee7c84ceee753eeaf795404d2b6374df501ab9ab2f02263f18e3604b6`
   - Restored the NicClusterPolicy to the original daemon tag `fix-sysfs-vpd-43a7ec3-vpd-debug-b84c98c-doca360-amd64`; both original-image pods are Ready with zero restarts and the policy is `ready`.
   - Restored the original launch-kit node labels and confirmed the temporary discovery namespace is absent.

No API/CRD schema changes were required.

almaslennikov:
Addressed the documentation finding and force-pushed amended commit `87086be`:

1. **Two-port north-south example:** changed it to `pfsPerNic: 2`, one representative PCI address, and `%plane_id%` prefixes so both PFs receive distinct `ns0`/`ns1` names.
2. **Placeholder guidance:** clarified in the README that `railPciAddresses` lists one address per `NicDevice`, `%nic_id%` is shared by its PFs, and `%plane_id%` provides distinct per-PF names.
3. **Scope:** left controller behavior unchanged; the existing whole-`NicDevice` ownership limitation remains documented separately.

Validation: the three-document YAML example parses successfully and `git diff --check` passes. No Go files changed in this review update.

PR #499, opened 2026-09-21
devicediscovery: fall back to sysfs for firmware info

### LINKED ISSUE #496: Discovery aborts on HGX B300: flint fails on onboard 4-port ConnectX-7 (MT2910), exposed by PCI-based device keying
## Summary

On NVIDIA HGX B300 nodes, `nic-configuration-daemon` fails to complete device discovery and enters a crash loop. Discovery aborts when it reaches the onboard 4-port ConnectX-7 (`MT2910`), because `GetFirmwareVersionAndPSID()` fails to run `flint` against it.

This is not a new defect, but it is newly *reachable*. It was previously masked by serial-number-based device grouping, and the switch to PCI-address-based keying (PR #351) exposes it. Any deployment that takes that fix on this hardware will hit it.

## Versions

| | |
|---|---|
| Fails | `nic-configuration-operator-daemon:network-operator-v26.4.2` |
| Works (masks the bug) | `nic-configuration-operator-daemon:v1.3.1` |
| Kubernetes | v1.35.3 |
| CRDs | from network-operator 26.1.0 |

Hardware: HGX B300, 8x single-port ConnectX-8 (`15b3:1023`) rails, 1x 4-port ConnectX-7 (`MT2910`), 1x BlueField-3 (`15b3:a2dc`).

## What happens

The daemon discovers the ConnectX-8 rails and the BlueField-3 successfully, then dies on the ConnectX-7:

```
INFO   devicediscovery.go:96  Found Mellanox device  {"address": "0000:0d:00.0", "type": "CX8 Family [ConnectX-8]"}
INFO   devicediscovery.go:96  Found Mellanox device  {"address": "0000:48:00.0", "type": "MT43244 BlueField-3 integrated ConnectX-7 network controller"}
INFO   devicediscovery.go:96  Found Mellanox device  {"address": "0000:a3:00.0", "type": "MT2910 Family [ConnectX-7]"}
ERROR  utils.go:202           GetFirmwareVersionAndPSID(): Failed to run flint  {"error": "exit status 1"}
ERROR  devicediscovery.go:125 Failed to get device's firmware and PSID          {"address": "0000:a3:00.0", "error": "exit status 1"}
ERROR  main.go:120            failed to discover NIC devices                    {"error": "exit status 1"}
```

The container then restarts and repeats. **No `NicDevice` CRs are written at all**, including for the eight ConnectX-8 rails that were discovered without error, because a single device failure aborts the whole pass.

Two observations that may help narrow it down:

- `GetVPD()` succeeds on the same ConnectX-7; only `flint` fails.
- The BlueField-3 (also ConnectX-7 silicon) is read successfully, firmware and PSID both returned. So this is specific to the standalone `MT2910`, not to ConnectX-7 generally.

## Why v1.3.1 does not hit it

Not because `flint` behaves differently. Because on this hardware `flint` is never invoked on that device.

The onboard NICs share a single flashed VPD image, so every port reports the same serial number. In v1.3.1, `GetFirmwareVersionAndPSID()` is called only inside the `if !ok` branch, when the serial is not yet present in the map:

```go
// pkg/devicediscovery/devicediscovery.go @ v1.3.1
deviceStatus, ok := devices[vpd.SerialNumber]

if !ok {
    firmwareVersion, psid, err := d.utils.GetFirmwareVersionAndPSID(device.Address)
    if err != nil {
        log.Log.Error(err, "Failed to get device's firmware and PSID", "address", device.Address)
        return nil, err
    }
```

The ConnectX-8 at the lowest bus is enumerated first and populates the entry. Every later port, including all four ConnectX-7 functions, matches the existing serial, skips the branch, and is never passed to `flint`.

After PR #351 the key is the PCI `Domain:Bus:Device`, so the ConnectX-7 gets its own key, takes the `if !ok` path for the first time, and the latent failure surfaces.

Worth noting the failure mode is identical in both versions once reached: v1.3.1 also does `return nil, err` here, so it would abort in exactly the same way.

## Workaround, and why it is not sufficient

Setting `SKIP_DEVICE_ON_DISCOVERY_ERROR=true` resolves the crash. Discovery completes, the ConnectX-7 is excluded, and everything else is processed correctly:

```
INFO  devicediscovery.go:92  Physical device already skipped during this discovery pass, skipping port  {"pciKey": "0000:a3:00", "address": "0000:a3:00.1"}
```

resulting in 9 `NicDevice` CRs (8 single-port ConnectX-8 + 1 BlueField-3), and `NicInterfaceNameTemplate` then applies correctly to all eight rails with `InterfaceNameApplied=True`.

This is fine when the unreadable card is one we do not need to manage. It does not generalise:

- The flag is all-or-nothing. There is no way to tolerate a failure on one device while still requiring the others, so a genuine failure elsewhere is silently downgraded to a skip.
- On a host where **every** NIC is a ConnectX-7, the same failure would exclude every device, leaving nothing for the operator to act on. That is a realistic configuration and blocks adopting the operator there at all.

## Questions

1. Is `flint` expected to work against the onboard 4-port `MT2910` on HGX B300? If it cannot be read, should discovery record the device without firmware/PSID rather than failing it?
2. Would you consider making a firmware/PSID read failure non-fatal by default, given it is only required for firmware management? Devices whose firmware cannot be read are still perfectly usable for interface naming and nvconfig.
3. Is a per-device or per-PCI-address skip list something you would accept, as a narrower alternative to the global flag?

Happy to gather any additional output from the affected hardware, including `flint` invoked directly with its stderr, if that would help.

almaslennikov:
## Problem

On HGX B300, the onboard four-port ConnectX-7 rejects firmware queries on PCI functions `.0` and `.1`: `flint` returns `ICMD bad parameter given`, while `devlink dev info` returns `Operation not supported`. Discovery queries `.0` first, so this aborts the entire pass even though the mlx5 driver exposes the required firmware metadata through RDMA sysfs on every function.

## Change

- keep `flint` as the primary firmware version and PSID source
- replace the devlink fallback with reads from the PCI function's RDMA `fw_ver` and `board_id` sysfs attributes
- require both values from the same RDMA device and preserve the existing discovery error and `SKIP_DEVICE_ON_DISCOVERY_ERROR` behavior when neither source is complete
- cover the reported ConnectX-7 values, a ConnectX-8 fixture, real sysfs symlink layout, incomplete attributes, and combined failures

No new mounts are required; the reporter confirmed these attributes are already readable from the daemon container.

Fixes #496

## Validation

```text
go test ./pkg/devicediscovery/... -v
go build ./...
make lint
```


almaslennikov (commented):
## Review verdict: no blocking findings

I reviewed the complete `8b4f0c6` change against issue #496, the reporter's diagnostics, and the repository conventions.

### Correctness

- The existing `flint` fast path remains first and unchanged in behavior.
- The fallback resolves the RDMA entry beneath the exact PCI function, so it does not borrow metadata from a sibling function.
- `fw_ver` and `board_id` must be non-empty on the same RDMA device; partial values from separate entries cannot be combined.
- Sysfs symlinks, multiple RDMA entries, empty attributes, attribute read failures, incomplete `flint` output, and combined `flint`/sysfs failure all have focused tests.
- Firmware version and PSID retain the prior lower-case normalization and the final error preserves both source failures.

### Security and performance

- The fallback is read-only and uses the daemon's existing `/sys` mount; it does not add capabilities, device access, subprocesses, resets, reloads, or writes.
- The implementation reads a small bounded directory and two small attributes per candidate, only after `flint` fails. I found no resource-lifetime or injection issue.

### Compatibility tradeoff

Greptile correctly notes that replacing devlink removes a fallback for a hypothetical system where RDMA sysfs metadata is absent but devlink metadata works. This is an intentional scope decision: on the reported hardware devlink has the same per-function restriction as `flint`, while the mlx5 RDMA sysfs attributes are available for every affected function and readable in the existing daemon container. The README now documents the resulting behavior. I do not consider this blocking for the requested change.

### Validation reviewed

- Local: `go test ./pkg/devicediscovery/... -v`, `go build ./...`, and `make lint` passed.
- Hosted CI: build, test, coverage, lint, go-check, CodeQL, license, DCO, and Greptile checks passed.
- Live cluster: a validation-only image forced `flint -d <pci> q` to fail. All four NICs on the healthy worker used sysfs successfully; discovered firmware/PSID values matched the raw `fw_ver`/`board_id` attributes, and the full eight-device inventory remained intact.
- Restoration readback: the original daemon image and policy are restored, the DaemonSet is 2/2 Ready, both workers are Ready and uncordoned, and both SR-IOV node states are `Succeeded`.

Because this is my own PR, GitHub does not allow me to submit an approving review; this `COMMENTED` review records the equivalent no-blocker verdict.

